PR #9, opened 2018-08-22
didChangeConfiguration Intercept

### LINKED ISSUE #10: Send Settings as part of Initialization Options
Ensure that settings are set in case of unexpected order of execution

### LINKED ISSUE #14: XML catalog settings
Please give the capability to configure teh XML catalog settings of lsp4xml language server with VSCode. Thanks!

NikolasKomonen:
Fixes #10 Fixes #9 Fixes #14

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,20 +60,30 @@
           "default": "off",
           "description": "Traces the communication between VS Code and the XML language server.",
           "scope": "window"
-        }
-      }
-    },
-    "configurationDefaults": {
-      "[xml]": {
-        "editor.insertSpaces": true,
-        "editor.tabSize": 2,
-        "editor.quickSuggestions": {
-          "other": true,
-          "comments": false,
-          "strings": true
         },
-        "editor.autoIndent": false,
-        "editor.autoClosingBrackets": false
+        "xml.catalogs": {
+          "type":"array",
+          "default": [],
+          "description": "Array of XML Catalogs",
+          "scope": "window"
+        },
+        "xml.logs.client": {
+         "type": "boolean",
+         "default": true,
+         "description": "Should the server log to client output"
+        },
+        "xml.format.splitAttributes": {
+          "type": "boolean",
+          "default": false,
+          "description": "Split multiple attributes each onto a new line",
+          "scope": "window"
+        },
+        "xml.format.joinCDATALines": {
+          "type": "boolean",
+          "default": false,
+          "description": "Preserve whitespace in tag with CDATA content",
+          "scope": "window"
+        }
       }
     }
   }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,17 +1,24 @@
 
-import * as vscode from "vscode";
 import { prepareExecutable } from './javaServerStarter';
-import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient } from 'vscode-languageclient';
+import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient, DidChangeConfigurationNotification } from 'vscode-languageclient';
 import * as requirements from './requirements';
-import { window, commands, ExtensionContext } from "vscode";
+import { workspace, window, commands, ExtensionContext } from "vscode";
 import * as path from 'path';
+import * as os from 'os';
+
+interface Settings {
+  catalogs: String[],
+  logs: {},
+  format: {}
+}
 
 export function activate(context: ExtensionContext) {
   let storagePath = context.storagePath;
   if (!storagePath) {
-    storagePath = "~/lsp4xml/";
+    storagePath = os.homedir() + "/.lsp4xml";
   }
   let logfile = path.resolve(storagePath + '/lsp4xml.log');
+
   return requirements.resolveRequirements().catch(error => {
     //show error
     window.showErrorMessage(error.message, error.label).then((selection) => {
@@ -26,11 +33,18 @@ export function activate(context: ExtensionContext) {
       // Register the server for java
       documentSelector: ['xml'],
       revealOutputChannelOn: RevealOutputChannelOn.Never,
-      initializationOptions: {"lsp4xml.logPath": logfile}
-      
+      initializationOptions: {settings: getSettings() },
+      synchronize: {
+        configurationSection: ['xml']
+      },
+      middleware: {
+        workspace: {
+          didChangeConfiguration: () => languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getSettings() })
+        }
+      }
+
     }
-    
-    
+
     let serverOptions = prepareExecutable(requirements);
     let languageClient = new LanguageClient('xml', 'XML Support', serverOptions, clientOptions);
     languageClient.onReady().then(() => {
@@ -38,7 +52,28 @@ export function activate(context: ExtensionContext) {
     });
     let disposable = languageClient.start();
     context.subscriptions.push(disposable);
-    
+
   });
+
+  function getSettings(): Settings {
+
+    let configXML = workspace.getConfiguration('xml');
+    let configCatalogs = configXML.get('catalogs') as String[];
+    let configFormats = configXML.get('format');
+
+    let configLogs = configXML.get('logs');
+    configLogs["file"] = logfile;
+
+    let settings: Settings = {
+      catalogs: configCatalogs,
+      logs: configLogs,
+      format: configFormats
+    }
+
+    return settings;
+  }
+
+
+
 
 }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -57,6 +57,56 @@
           "default": "off",
           "description": "Traces the communication between VS Code and the XML language server.",
           "scope": "window"
+        },
+        "xml.catalogs": {
+          "type":"array",
+          "default": [],
+          "description": "Catalog stuff",
```
fbricon: yes, but no

package.json
```diff
@@ -57,6 +57,56 @@
...
+          "scope": "window"
+        },
+        "xml.logs.client": {
```
fbricon: description?

package.json
```diff
@@ -57,6 +57,56 @@
...
+         "type": "boolean",
+         "default": true 
+        },
+        "xml.logs.file": {
```
fbricon: description?

fbricon: not sure we should actually expose that

src/extension.ts
```diff
@@ -26,11 +32,18 @@ export function activate(context: ExtensionContext) {
       // Register the server for java
       documentSelector: ['xml'],
       revealOutputChannelOn: RevealOutputChannelOn.Never,
-      initializationOptions: {"lsp4xml.logPath": logfile}
+      initializationOptions: {"lsp4xml.logPath": logfile},
```
fbricon: It might not be required now, but eventually you might want to send the settings as part of initializationOptions, if you need specific settings *before* handling the didChangeConfiguration request which is sent after the server has initialized.
Just keep that in mind for now.

src/extension.ts
```diff
@@ -41,4 +54,28 @@ export function activate(context: ExtensionContext) {
...
+    let configXML = workspace.getConfiguration('xml');
+    
+    let configCatalogs = configXML.get('catalogs') as String[];
+    
+    let configFormats = configXML.get('format');    
+    configFormats["indentUsingSpaces"] = workspace.getConfiguration("editor").get("insertSpaces");
+    configFormats["tabSize"] = workspace.getConfiguration("editor").get("tabSize");
+
+    let configLogs = configXML.get('logs');
+    if(configLogs['file'] == null) {
```
fbricon: should probably not be exposed to users. at least for now

src/extension.ts
```diff
@@ -41,4 +54,28 @@ export function activate(context: ExtensionContext) {
...
+      configLogs['file'] = logfile;
+    }
+
+    let settings: Settings = {
+      catalogs: configCatalogs,
+      logs: configLogs,
+      format: configFormats
+    }
+
+    return settings;
```
fbricon: doesn't the server expect all those settings to be prefixed with `lsp4xml.`?

src/javaServerStarter.ts
```diff
@@ -35,7 +35,7 @@ function prepareParams(requirements: RequirementsData): string[] {
     params.push('-jar'); params.push(path.resolve(server_home, launchersFound[0]));
   }else{
     return null;
-  }
```
fbricon: unnecessary (and bad) change

src/extension.ts
```diff
@@ -1,17 +1,23 @@
...
 import { prepareExecutable } from './javaServerStarter';
-import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient } from 'vscode-languageclient';
+import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient, DidChangeConfigurationNotification } from 'vscode-languageclient';
 import * as requirements from './requirements';
-import { window, commands, ExtensionContext } from "vscode";
+import { workspace, window, commands, ExtensionContext } from "vscode";
 import * as path from 'path';
 
+interface Settings {
+	catalogs: String[],
```
fbricon: indentation

src/extension.ts
NikolasKomonen: So I updated [the wiki](https://github.com/angelozerr/lsp4xml/wiki/Configuration) and tried to match the current layout of it and it seems like this is the nicest way to do it with the prefix being 'format'

package.json
NikolasKomonen: So only use the default path from extensions.js?

NikolasKomonen:
@fbricon updated

src/extension.ts
```diff
@@ -26,19 +32,67 @@ export function activate(context: ExtensionContext) {
...
+      insertSpaces = workspace.getConfiguration("editor").get("insertSpaces");
+    }
+    if (tabSize == null) {
+      tabSize = workspace.getConfiguration("editor").get("tabSize");
+    }
+    configFormats["insertSpaces"] = insertSpaces;
+    configFormats["tabSize"] = tabSize;
+
+
+    let configLogs = configXML.get('logs');
```
fbricon: should not be exposed to the client

src/extension.ts
```diff
@@ -26,19 +32,67 @@ export function activate(context: ExtensionContext) {
       // Register the server for java
       documentSelector: ['xml'],
       revealOutputChannelOn: RevealOutputChannelOn.Never,
-      initializationOptions: {"lsp4xml.logPath": logfile}
-      
+      initializationOptions: { "lsp4xml.logPath": logfile },
```
fbricon: all the settings should be sent in the initialization options, else logs.client is ignored on startup

package.json
```diff
@@ -57,6 +57,46 @@
...
+          "default": [],
+          "description": "Array of XML Catalogs",
+          "scope": "window"
+        },
+        "xml.logs.client": {
+         "type": "boolean",
+         "default": true,
+         "description": "Should the server log to client output"
+        },
+        "xml.logs.file": {
```
fbricon: remove from user-facing settings

src/extension.ts
```diff
@@ -1,17 +1,23 @@
...
+interface Settings {
+  catalogs: String[],
+  logs: {},
+  format: {}
+}
+
 export function activate(context: ExtensionContext) {
   let storagePath = context.storagePath;
   if (!storagePath) {
     storagePath = "~/lsp4xml/";
```
fbricon: `~` is not resolved. Fix it with:
```ts
import * as os from 'os';
...
storagePath = os.homedir()+"/.lsp4xml";
```

src/extension.ts
```diff
@@ -26,19 +32,67 @@ export function activate(context: ExtensionContext) {
...
+    }
+    if (tabSize == null) {
+      tabSize = workspace.getConfiguration("editor").get("tabSize");
+    }
+    configFormats["insertSpaces"] = insertSpaces;
+    configFormats["tabSize"] = tabSize;
+
+
+    let configLogs = configXML.get('logs');
+    if (configLogs['file'] == null) {
```
fbricon: should not be exposed to the client

fbricon:
This is blocked by https://github.com/angelozerr/lsp4xml/pull/94.
@angelozerr I suggest you merge 94 before your own changes, so we can move forward

package.json
```diff
@@ -60,20 +60,42 @@
...
+        },
+        "xml.logs.client": {
+         "type": "boolean",
+         "default": true,
+         "description": "Should the server log to client output"
+        },
+        "xml.format.splitAttributes": {
+          "type": "boolean",
+          "default": false,
+          "description": "Split multiple attributes each onto a new line",
```
fbricon: doesn't work

package.json
```diff
@@ -60,20 +60,42 @@
...
+          "scope": "window"
+        },
+        "xml.format.joinCDATALines": {
+          "type": "boolean",
+          "default": false,
+          "description": "Preserve whitespace in tag with CDATA content",
+          "scope": "window"
+        },
+        "xml.format.formatComments": {
```
fbricon: doesn't work

package.json
```diff
@@ -60,20 +60,42 @@
...
+          "type": "boolean",
+          "default": true,
+          "description": "Repositions comment tags and each line of content",
+          "scope": "window"
+        },
+        "xml.format.joinCommentLines": {
```
fbricon: doesn't work

NikolasKomonen:
@fbricon Should be good now

gulpfile.js
```diff
@@ -11,7 +11,7 @@ gulp.task('tslint', () => {
 });
 
 gulp.task('build_server', function(done) {
-	cp.execSync(mvnw()+ " clean verify -DskipTests", {cwd:server_dir, stdio:[0,1,2]} );
```
fbricon: please keep -DskipTests to make the build fast.

NikolasKomonen:
@fbricon Fixed the gulpfile

Issue #13, opened 2018-08-29
Improve detection of java executable


angelozerr:
In my case I have an installed java but not a JAVA_HOME. My java path is in my Windows PATH variable.

To use this java without configuring VSCode preferences, you could search it:

 * with `where java` command in OS Windows
 * with `which java` command for other OS.

This idea was done in Eclipse Bluesky to find the node process. Here the Java code:

```java
public static String getNodeJsLocation() {
		String res = "/path/to/node";
		String[] command = new String[] {"/bin/bash", "-c", "which node"};
		if (Platform.getOS().equals(Platform.OS_WIN32)) {
			command = new String[] {"cmd", "/c", "where node"};
		}
		BufferedReader reader = null;
		try {
			Process p = Runtime.getRuntime().exec(command);
			reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
			res = reader.readLine();
		} catch (IOException e) {
			Activator.getDefault().getLog().log(
					new Status(IStatus.ERROR, Activator.getDefault().getBundle().getSymbolicName(), e.getMessage(), e));
		} finally {
			IOUtils.closeQuietly(reader);
		}

		// Try default install path as last resort
		if (res == null && Platform.getOS().equals(Platform.OS_MACOSX)) {
			res = "/usr/local/bin/node";
		}

		if (res != null && Files.exists(Paths.get(res))) {
			return res;
		} else if (!alreadyWarned){
			warnNodeJSMissing();
			alreadyWarned = true;
		}
		return null;
	}
```

It should be cool to adapt this Java code in TypeScript and to discover java instead of node.

fbricon:
@angelozerr can you please try again with the latest build from http://download.jboss.org/jbosstools/vscode-xml/staging/?C=M;O=D. @snjeza checked this actually works on windows

angelozerr:
I have installed  vscode-xml-0.1.0-206.vsix but it doesn't work. I must define my java.home in preferences again.

fbricon:
where is your java directory located?
can you paste the content of %PATH%?

the java detection from the path is provided by https://github.com/redhat-developer/vscode-xml/blob/master/src/requirements.ts#L9 
Only way to debug it would be to sideload the vscode-xml extension, similar to [vscode-java](https://github.com/redhat-developer/vscode-java/blob/master/CONTRIBUTING.md#sideloading). (@NikolasKomonen will certainly add a contributing guide \*wink\* \*wink\*)

[...]

PR #33, opened 2018-09-13
 Support for JRE & Oracle (javapath instead of bin)

angelozerr:
This PR is just to show you  how I have fixed the search of java home with my Windows context:

 * no JAVA_HOME path
 * "where java" returns "C:\Program Files (x86)\Common Files\Oracle\Java\javapath\java.exe"
*  javapath is used insteaf of using bin

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -16,7 +16,7 @@ export function prepareExecutable(requirements: RequirementsData): Executable {
 	options.env = process.env;
 	options.stdio = 'pipe';
 	executable.options = options;
-	executable.command = path.resolve(requirements.java_home + '/bin/java');
+	executable.command = path.resolve(requirements.java_home + '/javapath/java');
 	executable.args = prepareParams(requirements);
 	return executable;
 }
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -60,7 +60,7 @@ function checkJavaRuntime(): Promise<string> {
             return resolve(javaHome);
         }
         //No settings, let's try to detect as last resort.
-        findJavaHome(function (err, home) {
+        findJavaHome({allowJre: true}, function (err, home) {
                 if (err){
                     openJDKDownload(reject,'Java runtime could not be located');
                 }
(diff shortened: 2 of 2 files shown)


GitMensch:
Coming here from #13 and wondering about the state of this PR - should the conflicts be fixed and the PR potentially be merged? Is that PR obsolete and should be closed instead?

GitMensch:
Note: I'm also on a machine where `where java` works (returns `C:\ProgramData\Oracle\Java\javapath\java.exe` here) but the plugin needs a manual setup. Should it?

fbricon:
This PR is stale. While the issue still exists, I think the best solution is to fix the detection mechanism upstream, or maybe, until this is fixed, reuse (and improve?) vscode-java's https://github.com/redhat-developer/vscode-java/blob/master/src/findJavaRuntimes.ts

GitMensch:
> I think the best solution is to fix the detection mechanism upstream, or maybe, until this is fixed, reuse (and improve?) vscode-java's https://github.com/redhat-developer/vscode-java/blob/master/src/findJavaRuntimes.ts

That's very reasonable! Do you see anything that needs improvement there (then this likely should be fixed in that upstream repo first before pulling it into this repo). The best option here, if that would work which I don't know,, would be to move findJavaRuntimes.ts to a separate node module which can then be used in both (and other) projects.

fbricon:
findJavaRuntime.ts doesn't support JRE detection at the moment, as vscode-java doesn't need it, so that definitely needs some improvements. Adding ./jenv/shim support would help bring feature parity with node-find-java-home (and adding sdkman support would be even better). @Eskibear any plans to publish it as a standalone npm module?

Eskibear:
I was planning to contribute to `node-find-java-home`, adding a new API to return all possible Java runtimes. But that would possibly break backward compatibility of the library, see https://github.com/jsdevel/node-find-java-home/issues/28#issuecomment-718818582 I'll discuss with the owner later, see if we can make some progress.

A standalone npm module can also be an alternative.

Eskibear:
I've published a npm package named [jdk-utils](https://www.npmjs.com/package/jdk-utils), which can list all installed Java runtimes, also coverring the Oracle Java SE 8 case.

image
image


Please give it a try and let me know if there's any case not coverred yet. See https://github.com/Eskibear/node-jdk-utils#findjavaruntime

GitMensch:
So that's closed - how was it solved?

PR #34, opened 2018-09-14
Register _xml.applyCodeAction to fix bug with CodeAction apply

angelozerr:
This PR register a custom command "_xml.applyCodeAction" to apply CodeAction returned by the XML Language Server.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2,7 +2,7 @@
 import { prepareExecutable } from './javaServerStarter';
 import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient, DidChangeConfigurationNotification } from 'vscode-languageclient';
 import * as requirements from './requirements';
-import { workspace, window, commands, ExtensionContext } from "vscode";
+import { workspace, window, commands, ExtensionContext, TextEdit } from "vscode";
 import * as path from 'path';
 import * as os from 'os';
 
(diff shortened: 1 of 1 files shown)


fbricon:
I think the server should send commands instead of the new textedit (protocol 3.8.0), so more clients are supported. Else you need to customize each client with that specific applyCodeAction command.

fbricon:
No. Server can only send CodeActions if the Client declares support for CodeAction literals, else it needs to fall back to returning command. That's why there are client capabilities

angelozerr:
> No. Server can only send CodeActions if the Client declares support for CodeAction literals, else it needs to fall back to returning command. That's why there are client capabilities

So if I understand your comment:

 * lsp4xml should return list of CodeAction (like before) according the client capabilities
 * vscode-xml should configure literal support capabilities. 

In lsp4j, we have:

```
CodeActionCapabilities codeActionCapabilities = new CodeActionCapabilities();
			codeActionCapabilities.setCodeActionLiteralSupport(new CodeActionLiteralSupportCapabilities());
```
But in typescript, I have not found this defintion? Could you tell me how to do that please.

fbricon:
No, the server must check if the client supports codeActions, by checking if capabilities.getTextDocument().getCodeAction().getCodeActionLiteralSupport() != null.
If codeActionLiteralSupport is supported, then return CodeActions. Else, return commands

fbricon:
codeActionLiteralSupport.codeActionKind.valueSet should not be empty

angelozerr:
> If codeActionLiteralSupport is supported, then return CodeActions. Else, return commands

Thanks for the feedback, done in https://github.com/angelozerr/lsp4xml/commit/89668441c8c6d8693744bb74ccfbfa7ae545a90d

> codeActionLiteralSupport.codeActionKind.valueSet should not be empty 

I have not implemented this test because LSP4e doesn't fill this codeActionLiteralSupport.codeActionKind.valueSet

fbricon:
This PR is made obsolete by #37. Updating the vscode client makes vscode-xml support codeActionLiterals, so CodeActions are now supported OOTB.

PR #35, opened 2018-09-14
Improve settings schema

felixfbecker:


package.json
```diff
@@ -93,7 +96,25 @@
...
+          "items": {
+            "type": "object",
+            "properties": {
+              "systemId": {
+                "type": "string",
+                "description": "The path or URL to the XML schema (XSD or DTD)"
+              },
+              "pattern": {
+                "type": "string",
+                "description": "File name pattern where you can use '*', '?' (it uses Java 7 glob NIO)"
```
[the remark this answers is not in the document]
felixfbecker: I experimented with something that works and added it as an example.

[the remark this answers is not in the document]
fbricon: not sure we can put a link in the descriptions here, but check https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-

package.json
```diff
@@ -93,7 +96,26 @@
...
+                "markdownDescription": "File glob pattern. Example: `**/*.Format.ps1xml`"
+              }
+            },
+            "required": [
+              "systemId",
+              "pattern"
+            ]
+          },
           "description": "Allows XML schemas to be associated to file name patterns.\n\nExample:\n[{\n  \"systemId\": \"path/to/file.xsd\",\n  \"pattern\": \"file1.xml\"\n},\n{\n  \"systemId\": \"path/to/file2.xsd\",\n  \"pattern\": \"file2.xml\"\n}]",
+          "markdownDescription": "Allows XML schemas to be associated to file name patterns.\n\nExample:\n```json\n[{\n  \"systemId\": \"path/to/file.xsd\",\n  \"pattern\": \"file1.xml\"\n},\n{\n  \"systemId\": \"path/to/file2.xsd\",\n  \"pattern\": \"file2.xml\"\n}]\n```",
```
fbricon: maybe use a glob pattern in file2 example

angelozerr: Perhaps you could add a link to https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob

fbricon: agreed, that's a better link

felixfbecker: Updated.

angelozerr: or perhaps this link https://github.com/angelozerr/lsp4xml/wiki/Configuration#file-associations

felixfbecker: I think it is good now

package.json
```diff
@@ -62,15 +62,18 @@
           "scope": "window"
         },
         "xml.catalogs": {
-          "type":"array",
+          "type": "array",
           "default": [],
+          "items": {
```
fbricon: "default": {
                "systemId": "path/to/file.xsd",
                "pattern": "**/*data.xml"
            },

felixfbecker: Why? Personally I would find that annoying as a user if I selected the autocompletion and it would prepopulate with an object with dummy values that have to manually delete and replace. It's much faster to use autcompletion to set the values, because it will generate empty strings and place your cursor into them.

Also, wouldn't `default` cause the vscode extension API to use that as a default value if it's not set?

fbricon: It's not the default value of xml.fileAssociations, I simply reproduced vscode's json.schemas [default item](https://github.com/Microsoft/vscode/blob/4419709b9bec8ad7cd1d385e6ec4dd7a13ab4bca/extensions/json-language-features/package.json#L39-L44).

Anyways,
```json
"default": {
            "systemId": "",
            "pattern": ""
        },
```
is a much better default than the current one.

felixfbecker: I don't understand the purpose.

As a user, this is how I add a config value:



This is what the `default` allows:



What is the point? What meaning has this "default value" as the hint says?
Using this method is a worse experience - I don't see the documentation for the `systemId` and `pattern` properties while editing.
I don't think it would be a better UX to not add this flag, as then users are always guided into the first flow.

fbricon: I liked having to the full skeleton on the 1st complete, instead of doing 3. But ok, I reckon the documentation part of the default value sucks. 
So let's do it your way for now, and see how users respond over time.

PR #36, opened 2018-09-17
Package.json updated for formatting capabilities, updated json Settings to include capabilities

NikolasKomonen:


src/extension.ts
```diff
@@ -10,7 +10,13 @@ interface Settings {
   catalogs: String[],
   logs: {},
   format: {},
-  fileAssociations: JSON[]
+  fileAssociations: JSON[],
+  serverCapabilities: {},
```
fbricon: why do you need to send serverCapabilities? There's no such preference

NikolasKomonen: Added that in by accident on a rebase

package.json
```diff
@@ -117,6 +117,12 @@
           "description": "Allows XML schemas to be associated to file name patterns.\n\nExample:\n[{\n  \"systemId\": \"path/to/file.xsd\",\n  \"pattern\": \"file1.xml\"\n},\n{\n  \"systemId\": \"http://www.w3.org/2001/XMLSchema.xsdd\",\n  \"pattern\": \"**/*.xsd\"\n}]",
           "markdownDescription": "Allows XML schemas to be associated to file name patterns.\n\nExample:\n```json\n[{\n  \"systemId\": \"path/to/file.xsd\",\n  \"pattern\": \"file1.xml\"\n},\n{\n  \"systemId\": \"http://www.w3.org/2001/XMLSchema.xsd\",\n  \"pattern\": \"**/*.xsd\"\n}]\n```",
           "scope": "window"
+        },
+        "xml.capabilities.formatting": {
```
fbricon: xml.format.enabled is more consistent

PR #37, opened 2018-09-18
Add xml logo and update vscode client

### LINKED ISSUE #20: CodeAction doesn't work in VSCode
I have tried to test code action in VSCode but it doesn't work altough with Eclipse LSP4E it works?

Here a sample:

```
<beans xmlns="http://www.springframework.org/schema/beans" xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<bean>
	                      dddddddddddddddddd
	</bean>
</beans>
```

Code action to remove dddddddddddddddddd is displayed, but when I click on it, it doesn't work? 

vscodecodeactionbug
Have you an idea? i sit a problem from lsp4xml? Thanks for your investigation.

fbricon:
Fixes #20


fbricon:
make sure to rm `-rf node_modules; npm install` after updating

PR #41, opened 2018-09-19
Changed setting from to xml.autoClosingTags.enabled to (xml.completion.autoCloseTags), now sends this info to server

NikolasKomonen:


NikolasKomonen:
@fbricon updated and tested

PR #56, opened 2018-10-04
Add XSL support

### LINKED ISSUE #55: Add  XSL support
Currently, when an xsl document is opened, it gets no XML support whatsoever, as it's set as its own language in vscode.
See [WhatIsXSL-example.xsl.zip](https://github.com/redhat-developer/vscode-xml/files/2444580/WhatIsXSL-example.xsl.zip) from https://www.w3.org/Style/XSL/WhatIsXSL.html

fbricon:
(Partially?) fixes #55

For some reason there's no autocompletion support in the xsl. @angelozerr do we need to modify the server as well?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,14 @@
   "bugs": "https://github.com/redhat-developer/vscode-xml/issues",
   "preview": true,
   "keywords": [
-    "xml"
+    "xml", "xsl", "xsd"
   ],
   "engines": {
     "vscode": "^1.27.0"
   },
   "activationEvents": [
-    "onLanguage:xml"
+    "onLanguage:xml",
+    "onLanguage:xsl"
   ],
   "main": "./out/src/extension",
   "scripts": {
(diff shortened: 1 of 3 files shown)


angelozerr:
The use of https://www.w3.org/1999/XSL/Transform don't give you the XML Schema, you have 3 XML Schema version for XSL. If you wish to support validation and completion for xsl, you need the my POC about XSL extension https://github.com/angelozerr/lsp4xml/tree/master/extensions/org.eclipse.lsp4xml.xsl but this https://github.com/angelozerr/lsp4xml/blob/master/extensions/org.eclipse.lsp4xml.xsl/src/main/java/org/eclipse/lsp4xml/xsl/XSLURIResolverExtension.java#L34 should be fixed (I think core should provide a folder where we can deploy some files of lsp4xml).

Today the extesnion doesn't use the version declared in the XML to choose the well XML Schema, I must do it.

fbricon:
@angelozerr I think xsl support should probably be a core feature as well

fbricon:
Regardless, I believe this PR is valuable on its own, even without the server components. @NikolasKomonen feel free to give it a try

angelozerr:
> @angelozerr I think xsl support should probably be a core feature as well

Ok, I will dot it. Please note it exists an issur for XSL at https://github.com/angelozerr/lsp4xml/issues/91

PR #57, opened 2018-10-04
Add default value if workspace configuration is not available

benoitf:
Theia is not implementing the workspace configuration for now (in progress)
So provide a default json value directly instead

Change-Id: Ie3e039954b2a0a20e128145794feaff4d2ea1c63
Signed-off-by: Florent BENOIT <fbenoit@redhat.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -89,8 +89,34 @@ export function activate(context: ExtensionContext) {
   function getSettings(): JSON {
     let configXML = workspace.getConfiguration();
     configXML = configXML.get('xml');
-    let x = JSON.stringify(configXML);
-    let settings: JSON = JSON.parse(x);
+    let settings: JSON;
+    if (!configXML) {
+      const defaultValue = {
+        logs: {
+
+        },
+        xml: {
+          trace: {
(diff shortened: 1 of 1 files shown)


fbricon:
Why is this needed? The server has the same default values so it should "just work"

benoitf:
@fbricon as I said in the description, Theia is not yet handling the workspace description so all settings of package.json are ignored when we deploy this vscode extension on it.
With that, it allows to make this extension for the meantime.

benoitf:
inside of Theia vscode-xml

fbricon:
I get that, but if no settings are sent, then the server's default settings will be used instead. AFAIK, the server's defaults are identical to vscode-xml's

benoitf:
so as `configXML` is undefined, and as code is using 
```typescript
settings['logs']['file'] = logfile;
```

you only want that I add an empty json supporting this attribute and not the others ?

NikolasKomonen:
@benoitf The structure of the json needs to be changed, but I'll submit an updated version of your pr. I've been testing it and it works well.

benoitf:
OK

PR #58, opened 2018-10-09
Add default value if workspace configuration is not available (modified)

NikolasKomonen:
Modification of #57


NikolasKomonen:
@fbricon can you review

PR #62, opened 2018-10-16
autoClosingBrackets are forced off if autoCloseTags is enabled

NikolasKomonen:


src/extension.ts
```diff
@@ -75,6 +75,17 @@ export function activate(context: ExtensionContext) {
 
   function getSettings(): JSON {
     let configXML = workspace.getConfiguration();
+
+    let autoCloseTags = configXML.get("xml.completion.autoCloseTags");
```
fbricon: this should only happen on configuration change on the client side.
there's no need to send that config to the server

src/extension.ts
```diff
@@ -75,6 +75,17 @@ export function activate(context: ExtensionContext) {
...
+    let autoClosingBrackets = configXML.get("[xml]")["editor.autoClosingBrackets"];
+    if (autoCloseTags && autoClosingBrackets != "never") {
```
fbricon: you need to ask permission for changing the value. Not do it 1st and ask for forgiveness

src/extension.ts
```diff
@@ -45,11 +44,12 @@ export function activate(context: ExtensionContext) {
       revealOutputChannelOn: RevealOutputChannelOn.Never,
       initializationOptions: { settings: getSettings() },
       synchronize: {
-        configurationSection: ['xml']
+        configurationSection: ['xml', '[xml]']
```
fbricon: is this required? The server doesn't need to get synchronize on [xml] I think

NikolasKomonen: It allows for [specific language configurations](https://github.com/redhat-developer/vscode-xml/pull/62/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R155) to be considered for didChangeConfiguration.

src/extension.ts
```diff
@@ -102,11 +105,33 @@ export function activate(context: ExtensionContext) {
...
+      "'xml.completion.autoCloseTags' cannot be enabled while '[xml].editor.autoClosingBrackets' does not equal 'never'.", 
+        "autoCloseTags = false", 
+        "autoClosingBrackets = 'never'").then((selection) => {
+          if(selection == "autoClosingBrackets = 'never'") {
+            workspace.getConfiguration().update("[xml]", { "editor.autoClosingBrackets": "never" }, true).then(
+              () => console.log('[xml].editor.autoClosingBrackets globally set to never'),
+              (error) => console.log(error)
+            );
+          }
+          else {
```
fbricon: be careful, this is the case when user presses escape. You don't want to do anything if action gets cancelled on purpose

NikolasKomonen: good catch, ill fix

src/extension.ts
```diff
@@ -102,11 +105,33 @@ export function activate(context: ExtensionContext) {
...
 
-
-
-
+function verifyConfigurations() {
+  let configXML = workspace.getConfiguration();
+  let autoCloseTags = configXML.get("xml.completion.autoCloseTags");
+  let autoClosingBrackets = configXML.get("[xml]")["editor.autoClosingBrackets"];
+  if (autoCloseTags && autoClosingBrackets != "never") {
+    window.showInformationMessage(
```
fbricon: showWarningMessage

src/extension.ts
```diff
@@ -102,11 +105,33 @@ export function activate(context: ExtensionContext) {
...
 
-
-
-
+function verifyConfigurations() {
+  let configXML = workspace.getConfiguration();
+  let autoCloseTags = configXML.get("xml.completion.autoCloseTags");
+  let autoClosingBrackets = configXML.get("[xml]")["editor.autoClosingBrackets"];
+  if (autoCloseTags && autoClosingBrackets != "never") {
+    window.showWarningMessage(
```
fbricon: the message briefly shows up when the extension starts, then disappears. You should probably store a copy of the config on each check. When the 1st check happens, there's no oldConfig value, you do nothing (except storing the current config into oldConfig)

As for the message, I'd rather display something like:
> "The `[xml].editor.autoClosingBrackets` setting conflicts with `xml.completion.autoCloseTags`. It's recommended to disable it."

with 2 options: `Disable` and `Don't show again`.
If disable is selected, you need to check where is autoClosingBrackets set, using configXML.inspect(key). Then update the new value at the same place (globally vs locally)
If don't show again, then you need to store a `xml.completion.autoClosingBracketsConflictWarning` globally, that you can check again on next call to verifyConfigurations()

NikolasKomonen: The message disappearing is because of didChangeConfiguration being called twice since we have intializationOptions being called (I dont think its needed anyways).

And Ill fix to match what you said for the message.

src/extension.ts
```diff
@@ -102,11 +117,64 @@ export function activate(context: ExtensionContext) {
...
     return settings;
   }
+}
 
 
+function verifyAutoClosing() {
+  let configXML = workspace.getConfiguration();
+  let autoCloseTags = configXML.get("xml.completion.autoCloseTags");
+  let autoClosingBrackets = configXML.get("[xml]");
+  console.log(autoClosingBrackets);
```
fbricon: remove log

src/extension.ts
```diff
@@ -102,11 +117,64 @@ export function activate(context: ExtensionContext) {
...
+  autoClosingBrackets = autoClosingBrackets["editor.autoClosingBrackets"];
```
fbricon: please use a different variable name. this is too confusing

package.json
```diff
@@ -150,6 +150,11 @@
           "scope": "window"
```
fbricon: Please change the description to "Enable/disable autoclosing of XML tags. \n\nIMPORTANT: Turn off editor.autoClosingTags for this to work", to make it consistent with other "autoclosing" preferences in VS Code.

fbricon:
Few minor changes left to do but it looks pretty good now.

NikolasKomonen:
@fbricon Updated

PR #97, opened 2019-01-14
Modified JSON preferences structure

NikolasKomonen:
Another layer to preferences was added ('xml') which encapsulates the existing preferences object.

Related to: https://github.com/angelozerr/lsp4xml/issues/257

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -52,7 +52,8 @@ export function activate(context: ExtensionContext) {
       // Register the server for xml and xsl
       documentSelector: ['xml', 'xsl'],
       revealOutputChannelOn: RevealOutputChannelOn.Never,
-      initializationOptions: { settings: getSettings() },
+      //wrap with key 'settings' so it can be handled same a DidChangeConfiguration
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -111,14 +120,15 @@ export function activate(context: ExtensionContext) {
         }
       }
       const x = JSON.stringify(defaultValue);
-      settings = JSON.parse(x);
+      xml = JSON.parse(x);
     } else {
       const x = JSON.stringify(configXML);
-      settings = JSON.parse(x);
+      xml = JSON.parse(x);
```
fbricon: those 2 lines should move outside the if/else block

src/extension.ts
```diff
@@ -111,14 +120,15 @@ export function activate(context: ExtensionContext) {
         }
       }
       const x = JSON.stringify(defaultValue);
-      settings = JSON.parse(x);
+      xml = JSON.parse(x);
```
fbricon: you're missing the `xml = { "xml" : xml};` bit here. Anyways line should be removed because of previous comment

PR #100, opened 2019-01-23
Preference to enable validation

NikolasKomonen:
In relation to https://github.com/angelozerr/lsp4xml/issues/260

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
           "description": "Specifies extra VM arguments used to launch the XML Language Server. Eg. use `-noverify -Xmx1G  -XX:+UseG1GC -XX:+UseStringDeduplication` to bypass class verification, increase the heap size to 1GB and enable String deduplication with the G1 Garbage collector",
           "scope": "window"
         },
-        "xml.problems.noGrammar": {
+        "xml.validation.noGrammar": {
           "type": "string",
           "enum": [
             "ignore",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -176,6 +176,19 @@
...
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable all diagnostics.",
+          "scope": "window"
+        }
+        ,
+        "xml.diagnostics.schema": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable schema based diagnostics.",
```
fbricon: Enable/disable schema based diagnostics. Ignored if `"xml.diagnostics.enabled":false`

package.json
```diff
@@ -176,6 +176,19 @@
...
+          "scope": "window"
+        }
+        ,
+        "xml.diagnostics.schema": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable schema based diagnostics. Ignored if \"xml.diagnostics.enabled\": false.",
+          "scope": "window"
         }
       }
```
fbricon: I'm thinking we need a consistent namespace for everything, eg. xml.validation.
so:
- xml.validation.enabled
- xml.validation.schema.enabled
- xml.validation.noGrammar
we'd have to deprecate xml.problems.noGrammar

NikolasKomonen:
@fbricon updated with diagnostics -> validation

PR #116, opened 2019-03-19
Exclude ParentProcessWatcher by default

ralish:
Disables the parent process watcher for lsp4xml by default.

For additional context on this change see: https://github.com/angelozerr/lsp4xml/issues/328.

The support in lsp4xml is introduced in: https://github.com/angelozerr/lsp4xml/pull/330.

This change is effectively mirroring that introduced to vscode-java by: https://github.com/redhat-developer/vscode-java/pull/843.

### DIFF
--- a/src/javaServerStarter.ts
+++ b/src/javaServerStarter.ts
@@ -1,6 +1,7 @@
 import { workspace } from 'vscode'
 import { Executable, ExecutableOptions } from 'vscode-languageclient';
 import { RequirementsData } from './requirements';
+import * as os from 'os';
 import * as path from 'path';
 const glob = require('glob');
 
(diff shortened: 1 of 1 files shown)


NikolasKomonen:
I think we need to set defaults specific to the OS. I've tested on Fedora and the server does not end if the client does not close properly.

I've tested on Windows and it looks like the PPW isn't necessary there.

MacOS hasn't been tested yet.

@fbricon would you be able to test MacOS when you have time?

fbricon:
@NikolasKomonen how did you check on Fedora. @snjeza said should couldn't get zombie processes on linux. It was fine on Mac, as far as I tested

fbricon:
@ralish can you please sign-off your commit? (git commit -s)

NikolasKomonen:
@fbricon

To recreate it I:

1) I ran `jps` to see if the `...uber.jar` existed.
2) I ran `pkill code` to kill all instances of VSCode
3) I waited a bit then ran `jps`, vscode-java disappeared but vscode-xml was still there.

fbricon:
@ralish please do the same thing as in https://github.com/redhat-developer/vscode-java/pull/847/files#diff-d106f0e1914f2f6cbf79404a7f0c42b4R68

ralish:
@fbricon: Amended commit per `vscode-java` implementation and signed as requested.

PR #121, opened 2019-03-27
Configuration for storageDirectory

### LINKED ISSUE #120: ".lsp4xml" directory should be configurable
In reference to https://github.com/angelozerr/lsp4xml/issues/222

NikolasKomonen:
Fixes #120

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -34,7 +34,13 @@ let ignoreVMArgs = false;
 export function activate(context: ExtensionContext) {
   let storagePath = context.storagePath;
   if (!storagePath) {
-    storagePath = os.homedir() + "/.lsp4xml";
+    let preferencePath : string = workspace.getConfiguration().get('xml.storagePath');
+    if(preferencePath && preferencePath.length > 0) {
+      storagePath = preferencePath;
+    }
+    else {
+      storagePath = os.homedir() + "/.lsp4xml";
+    }
   }
   let logfile = path.resolve(storagePath + '/lsp4xml.log');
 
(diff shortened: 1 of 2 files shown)


fbricon:
What we want is to expose the server's lsp4xml.workdir property through preferences (could be renamed to `xml.server.workDir`). So this also requires server side changes: https://github.com/angelozerr/lsp4xml/blob/a77306975b2312a1a219789a8d437d44064e59d8/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java#L48

src/extension.ts
```diff
@@ -34,7 +34,13 @@ let ignoreVMArgs = false;
...
-    storagePath = os.homedir() + "/.lsp4xml";
+    let preferencePath : string = workspace.getConfiguration().get('xml.storagePath');
+    if(preferencePath && preferencePath.length > 0) {
+      storagePath = preferencePath;
+    }
+    else {
+      storagePath = os.homedir() + "/.lsp4xml";
+    }
   }
   let logfile = path.resolve(storagePath + '/lsp4xml.log');
```
fbricon: the problem here is we need to separate the logs from the workspace (which belongs to context.storagePath), from the caching directory reusable by all lsp4xml instances (a maven schema, once downloaded for a particular directory, can be reused by another)

NikolasKomonen:
Ignore this PR, wrong Idea.

Proper implementation by https://github.com/redhat-developer/vscode-xml/pull/123

PR #133, opened 2019-04-30
Preserve whitespace on format

### LINKED ISSUE #128: Preserve blank lines
Often, it is better to read an XML document with major elements separated with an empty line. Please provide an option to preserve blank lines.

Example of expected code to be formatted, but keeping empty lines:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<project
        xmlns="http://maven.apache.org/POM/4.0.0"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

    <modelVersion>4.0.0</modelVersion>
    <groupId>com.foo</groupId>
    <artifactId>bar</artifactId>
    <version>1.0-SNAPSHOT</version>
    <packaging>jar</packaging>

    <name>Foo</name>
    <description>Foo project</description>

    <properties>
        ...
    </properties>

    <dependencyManagement>
        <dependencies>
            ...
        </dependencies>
    </dependencyManagement>

    <build>
        <pluginManagement>
           ...
        </pluginManagement>

        <plugins>
           ...
        </plugins>
    </build>

    <reporting>
       ...
    </reporting>

    <modules>
       ...
    </modules>
</project>
```

NikolasKomonen:
Fixes #128

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,18 @@
           "description": "Preserve empty content/whitespace in a tag.",
           "scope": "window"
         },
+        "xml.format.preservedNewlines": {
+          "type": "integer",
+          "enum": [
+            0,
+            1,
+            2,
+            3
+          ],
+          "default": 2,
+          "description": "Preserve new lines that separate tags. The value represents the maximum number of new lines per section. A value of 0 removes all new lines.",
+          "scope": "window"
+        },
         "xml.format.enabled": {
           "type": "boolean",
           "default": true,


package.json
```diff
@@ -124,6 +124,12 @@
           "description": "Preserve empty content/whitespace in a tag.",
           "scope": "window"
         },
+        "xml.format.preserveNewlines": {
+          "type": "boolean",
+          "default": true,
+          "description": "Preserve newlines that separate tags. If the amount of new lines exceeds 2 it will be reduced to 2, and everything under is preserved.",
```
fbricon: "Preserve new lines"

package.json
```diff
@@ -124,6 +124,12 @@
           "description": "Preserve empty content/whitespace in a tag.",
           "scope": "window"
         },
+        "xml.format.preserveNewlines": {
+          "type": "boolean",
```
fbricon: maybe we should just authorize int values instead:  [0 ,1, 2, 3]. Set default to 2, 0 means no line preserved

package.json
```diff
@@ -124,6 +124,18 @@
           "description": "Preserve empty content/whitespace in a tag.",
           "scope": "window"
         },
+        "xml.format.preserveNewlines": {
```
fbricon: xml.format.preservedNewlines see https://github.com/angelozerr/lsp4xml/pull/364/files#r280883994

fbricon:
Funny enough, any value that's outside the 0...4 boundaries is still accepted and respected on the server side. But it's fine, that's not a concern.

PR #136, opened 2019-05-15
JRE support

angelozerr:
Only JDK is supported. Generally users who don't develop with Java, have an installed JRE not a JDK. I think javac is used here because it was a copy/paste from jdt ls code which requires JDK.

This very simple PR checks if java exists and not javac. But it requires to declare the JRE with java.home. It should be improved to find JRE like we have that with JDK.


fbricon:
We need to introduce a vscode-xml specific java.home variable, maybe `xml.java.home`, that supports a JRE, that'd be checked 1st, then fall back to java.home (which, when used along vscode-java needs to be a JDK)

fbricon:
Closing as superseded by #152

PR #143, opened 2019-05-24
Fix bad indentation issue

### LINKED ISSUE #137: The plugin does not properly indent XML
Hey,
I am using the version 0.5.1 of the plugin and noticed something really strange. 
The indentation upon hitting 'enter' on some text does not work as expected:

I have the following code (the pipe is the cursor)
``` 
  <para>
    A disabled service will not be started automatically.|</para>
```
After hitting enter, here's the result:
``` 
  <para>
    A disabled service will not be started automatically
    </para>
```

Should be:
``` 
  <para>
    A disabled service will not be started automatically
  </para>
```

If I use HTML as the file type in VSCode, indentation works as expected.

xorye:
Fixes #137

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -212,6 +212,13 @@ function getScopeLevel(configurationKey : string, key : string) : ScopeInfo{
 
 function getIndentationRules(): LanguageConfiguration {
   return {
+    
+    // indentationRules referenced from:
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -212,6 +212,10 @@ function getScopeLevel(configurationKey : string, key : string) : ScopeInfo{
 
 function getIndentationRules(): LanguageConfiguration {
   return {
```
fbricon: please add a comment linking to the source of the regexps

xorye: I added the comment. How is it?

PR #152, opened 2019-06-05
JRE support

xorye:
JRE path can now be detected automatically. This was tested on a machine with JRE, but without JDK.

This PR branch was initially taken from #136 . 

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,13 +36,14 @@ See the [changelog](CHANGELOG.md) for the latest release. You might also find us
 
 ## Requirements
 
-  * Java JDK 8 or more recent
+  * Java JDK (or JRE) 8 or more recent
   * Ensure Java path is set in either: 
     * `xml.java.home` in VSCode preferences
     * `java.home` in VSCode preferences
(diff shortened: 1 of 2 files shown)


fbricon:
Please also update the readme

xorye:
Updated. Does this suffice?

xorye:
On second thought, I will add some more information.

fbricon:
OK this work but i found an issue with the code you changed from https://github.com/redhat-developer/vscode-xml/commit/ef91528ec95fab89626633b8657303d2180d741a#diff-baff02b04010a735985f63100efb9fc2L37:

each case is checked even after the java home path has been found. For some reason, it doesn't cause an error because apparently the 1st call to `resolve` wins. The old code, albeit uglier, used to better handle one case after the other

xorye:
Yes, that was one of the benefits of the previous version. It turns out that when `resolve()` is called, it still executes everything else after it, but the first resolve is always the result that the promise returns. Do you want me to change it back to the previous way?

fbricon:
Yes please

xorye:
I have reverted it

PR #158, opened 2019-06-18
Restart message when a java path preference is changed

### LINKED ISSUE #150: JDK changes (xml.java.home or java.home) should be detected
... and user should be prompted to restart vscode (similar to xml.server.vmargs and vscode-java)

NikolasKomonen:
Fixes #150

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -1,0 +1,146 @@
+import { WorkspaceConfiguration, workspace, window, commands, extensions, Extension } from "vscode";
+import { ScopeInfo } from "./extension";
+
+let vmArgsCache;
+let ignoreAutoCloseTags = false;
+let ignoreVMArgs = false;
+let oldXMLConfig: WorkspaceConfiguration = getXMLConfiguration();
+let oldJavaConfig: WorkspaceConfiguration = getJavaConfiguration();
+
+const restartButton = 'Restart Now';
+const ignoreButton = 'Ignore'
+const restartId = "workbench.action.reloadWindow";
+
+
+export function getXMLConfiguration(): WorkspaceConfiguration {
+  return getXConfiguration("xml")
+}
+
+export function getJavaConfiguration(): WorkspaceConfiguration {
+  return getXConfiguration("java");
+}
+
+export function getXConfiguration(configName: string) {
+  return workspace.getConfiguration(configName);
+}
+
+export function onConfigurationChange() {
+  if(!ignoreVMArgs) {
+    verifyVMArgs();
+  }
+
+  if(!ignoreAutoCloseTags) {
+    verifyAutoClosing();
+  }
+}
+
+export function subscribeJDKChangeConfiguration() {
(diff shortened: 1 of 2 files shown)


src/settings.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    const newConfig = getXMLConfiguration();
+
+    if(hasPreferenceChanged(oldConfig, newConfig, "java.home")) { // checks "xml.java.home"
+      const message = "`xml.java.home` path has changed. Please restart VS Code."
+      createReloadWindowMessage(message);
+    }
+
+    else if(javaHomeConfig != getJavaHomeConfiguration()) { // checks "java.home"
```
fbricon: on check it if xml.java.home is not set, else you can ignore the change, it doesn't impact XML

fbricon:
the change should be refactored along with verifyVMArgs. Now we have 2 change listeners

NikolasKomonen:
@fbricon updated, moved into 1 listener in the settings.ts file

src/settings.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export function onConfigurationChange() {
+  if(!ignoreAutoCloseTags) {
+    verifyAutoClosing();
+  }
+  
+  !ignoreVMArgs ? verifyVMArgs(): undefined;
+
+  workspace.onDidChangeConfiguration(params => {
```
fbricon: this seems wrong. This needs to be registered once. Calling it from inside middleware.workspace.didChangeConfiguration doesn't seem right. There must be a way to get the change event directly from middleware.workspace.didChangeConfiguration

fbricon:
@snjeza do you mind looking into this one, while @NikolasKomonen is away?

src/settings.ts
```diff
@@ -0,0 +1,122 @@
...
+export function getXMLConfiguration(): WorkspaceConfiguration {
+  return workspace.getConfiguration("xml");
+}
+
+export function onConfigurationChange() {
+  if(!ignoreAutoCloseTags) {
+    verifyAutoClosing();
+  }
+  
+  !ignoreVMArgs ? verifyVMArgs(): undefined;
```
fbricon:
```typescript
  if (!ignoreVMArgs) {
    verifyVMArgs();
  }
```

NikolasKomonen:
@fbricon Updated, setup the jdk listener to subscribe only once

It starts at this line: https://github.com/redhat-developer/vscode-xml/blob/a190861729b680bc04da2e67cef58e823a608358/src/extension.ts#L94

NikolasKomonen:
@fbricon 

Updated,

The prompt should appear when the 'xml.java.home' setting is changed.
If 'java.home' is changed but 'xml.java.home' exists, then nothing happens.
If 'java.home' changes, 'xml.java.home' doesn't exist, and the Java extension is inactive/missing, then you get the prompt.

PR #164, opened 2019-07-03
XSD CodeLens references should be clickable

angelozerr:
See https://github.com/angelozerr/lsp4xml/issues/490

This PR binds the XSD CodeLens references command to a function which
opens the vscode references.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -1,0 +1,33 @@
+/**
+ *  Copyright (c) 2019 Red Hat, Inc. and others.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v1.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v10.html
+ *
+ *  Contributors:
+ *  Red Hat Inc. - initial API and implementation
+ */
+'use strict';
+
+/**
+ * Commonly used commands
+ */
+export namespace Commands {
+ 
+    /**
+     * Auto close tags
+     */
+    export const AUTO_CLOSE_TAGS= 'xml.completion.autoCloseTags';
+
+    /**
+     * Show XML references
+     */
+    export const SHOW_REFERENCES = 'xml.show.references';
+
+    /**
+     * Show editor references
+     */
+    export const EDITOR_SHOW_REFERENCES = 'editor.action.showReferences';
+
+}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -11,12 +11,13 @@
  */
 
 import { prepareExecutable } from './javaServerStarter';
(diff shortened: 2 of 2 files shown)


src/extension.ts
```diff
@@ -83,14 +95,27 @@ export function activate(context: ExtensionContext) {
     let disposable = languageClient.start();
     toDispose.push(disposable);
     languageClient.onReady().then(() => {
+
+      // Code Lens actions
+      context.subscriptions.push(commands.registerCommand(Commands.SHOW_XML_REFERENCES, (uriString: string, position: Position) => {
```
fbricon: SHOW_REFERENCES, else it doesn't compile

src/extension.ts
```diff
@@ -83,14 +95,27 @@ export function activate(context: ExtensionContext) {
...
+        const uri = Uri.parse(uriString);
+        workspace.openTextDocument(uri).then(document => {
+          // Consume references service from the XML Language Server
+          let param = languageClient.code2ProtocolConverter.asTextDocumentPositionParams(document, position);
+          languageClient.sendRequest(ReferencesRequest.type, param).then(locations => {
+            commands.executeCommand(Commands.SHOW_REFERENCES, uri, languageClient.protocol2CodeConverter.asPosition(position), locations.map(languageClient.protocol2CodeConverter.asLocation));
```
fbricon: EDITOR_SHOW_REFERENCES or else you end up in an infinite loop

src/commands.ts
```diff
@@ -0,0 +1,33 @@
...
+ 
+    /**
+     * Auto close tags
+     */
+    export const AUTO_CLOSE_TAGS= 'xml.completion.autoCloseTags';
+
+    /**
+     * Show LSP Standard references
+     */
+    export const SHOW_REFERENCES = 'lsp.show.references';
```
fbricon: there's no such thing as LSP standard references. We should keep xml.show.references until Reference Code lenses are defined at the protocol level.

src/extension.ts
angelozerr: Sorry for this big mistake, @fbricon fixed.

src/commands.ts
angelozerr: @fbricon updated

src/extension.ts
angelozerr: Sorry for this big mistake, @fbricon fixed.

src/commands.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export namespace Commands {
+ 
+    /**
+     * Auto close tags
+     */
+    export const AUTO_CLOSE_TAGS= 'xml.completion.autoCloseTags';
+
+    /**
+     * Show LSP Standard references
```
fbricon: Show XML references

angelozerr:
Thanks @fbricon !

PR #165, opened 2019-07-05
Add 'xml.telemetry.enabled' settings

angelozerr:
Fix https://github.com/angelozerr/lsp4xml/issues/430

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ The following settings are supported:
 Since 0.8.0:
 * `xml.codeLens.enabled`: Enable/disable XML CodeLens. Default is `false`.
 * `xml.symbols.excluded`: Disable document symbols (Outline) for the given file name patterns. Updating file name patterns does not automatically reload the Outline view for the relevant file(s). Each file must either be reopened or changed, in order to trigger an Outline view reload.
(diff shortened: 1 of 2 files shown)


fbricon:
We're not gonna enable telemetry settings in vscode until we can actually collect the data

angelozerr:
> We're not gonna enable telemetry settings in vscode until we can actually collect the data

You mean that vscode-xml must not provide `xml.telemetry.enabled `settings?  If it that it means that it's a more client capability?

fbricon:
Please change the commit message as this doesn't "fix" the lsp4xml issue. Use "relates to" instead

fbricon:
Actually we don't need this PR for now.

angelozerr:
> Actually we don't need this PR for now.

I don't understand, how can we enable telemetry in this case since telemetry is disabled by default?

fbricon:
we can still enable telemetry for testing purposes by adding xml.telemetry.enabled:true to the settings. It won't appear as a valid preference, but we power users will know ;-)

The preference will be exposed once vscode actually collect data

angelozerr:
Ok thank's for the clarification @fbricon

PR #193, opened 2019-10-22
Contribute to completion, hover .. with external JAR

### LINKED ISSUE #190: Contribute to completion, hover .. with external JAR
Liferay is working on LSP4XML extension.

maven extension https://github.com/angelozerr/lsp4xml/issues/166 is an another LSP4XML extension.

To support those extensions, vscode-xml should provide a mechanism to declare in the package JSON of the project the JAR which contains extension registered with SPI:

```json
"contributes": {
		"xml.javaExtensions": [
			"./jar/org.eclipse.lsp4xml.extensions.maven-0.9.1.jar"
		]
}

```
When server is started it loop for each JAR to include it in the classpath (like vscode-java does).

angelozerr:
Contribute to completion, hover .. with external JAR

Fixes #190

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "dtd"
   ],
   "engines": {
-    "vscode": "^1.27.0"
+    "vscode": "^1.37.0"
   },
   "activationEvents": [
     "onLanguage:xml",
@@ -28,10 +28,11 @@
     "compile": "webpack --mode none",
     "watch": "webpack --mode development --watch --info-verbosity verbose",
     "test-compile": "tsc -p ./",
-    "postinstall": "node ./node_modules/vscode/bin/install",
-    "test": "node ./node_modules/vscode/bin/test",
-    "build-server": "./node_modules/.bin/gulp build_server",
-    "tslint": "./node_modules/.bin/gulp tslint"
+    "pretest": "npm run test-compile",
+    "test": "node ./out/test/runTest.js",
+    "tslint": "./node_modules/.bin/tslint -p .",
+    "build": "./node_modules/.bin/gulp build",
+    "build-server": "./node_modules/.bin/gulp build_server"
   },
   "repository": {
     "type": "git",
@@ -44,25 +45,27 @@
     "Snippets"
   ],
   "devDependencies": {
-    "@types/mocha": "^5.2.5",
-    "@types/node": "^6.0.40",
-    "gulp": "^4.0.0",
+    "@types/mocha": "^5.2.6",
+    "@types/node": "^10.14.16",
+    "@types/vscode": "^1.37.0",
+    "gulp": "^4.0.2",
     "gulp-tslint": "^8.1.3",
-    "mocha": "^5.2.0",
-    "tar": ">=4.4.2",
+    "gulp-rename": "^1.4.0",
+    "mocha": "^6.2.1",
     "ts-loader": "^6.0.1",
-    "tslint": "^5.11.0",
-    "typescript": "^3.0.3",
-    "vscode": "^1.1.21",
+    "tslint": "^5.12.1",
+    "typescript": "^3.4.1",
+    "typescript-tslint-plugin": "^0.3.1",
+    "vscode-test": "^1.2.0",
     "webpack": "^4.32.2",
     "webpack-cli": "^3.3.2"
   },
   "dependencies": {
     "expand-home-dir": "^0.0.3",
     "find-java-home": "0.2.0",
-    "glob": "^7.1.1",
-    "path-exists": "^3.0.0",
-    "vscode-languageclient": "~5.1.0"
+    "glob": "^7.1.4",
+    "path-exists": "^4.0.0",
+    "vscode-languageclient": "^5.2.1"
   },
   "contributes": {
     "configuration": {
@@ -284,6 +287,12 @@
           ".ent",
           ".mod"
         ]
+      }
+    ],
+    "jsonValidation": [
+      {
+        "fileMatch": "package.json",
+        "url": "./schemas/package.schema.json"
       }
     ]
   }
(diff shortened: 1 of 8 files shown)


angelozerr:
This PR gives the capability to contribute with custom completion, hover, etc by using SPI mechanism used by LSP4XML by developping a vscode extension which declares in the package.json, the xml.javaExtensions:

```json
       "contributes": {
		"xml.javaExtensions": [
			"./jar/org.eclipse.lsp4xml.extensions.maven-0.9.1.jar"
		]
	}
```

See a sample at https://github.com/angelozerr/vscode-xml-maven/blob/master/package.json#L18

Please review the code for the moment and I will explain you soon how to test it.

angelozerr:
To test this PR you can do it with https://github.com/angelozerr/vscode-xml-maven/blob/master/package.json#L18

This vscode extension extends the completion scope for maven by using LSP5XML SPI completion participant for maven. See https://github.com/angelozerr/lsp4xml-extensions-maven/blob/master/org.eclipse.lsp4xml.extensions.maven/src/main/java/org/eclipse/lsp4xml/extensions/maven/MavenCompletionParticipant.java#L39

As we cannot launch 2 extensions in vscode, you should 

 * install [vscode-xml-maven-0.0.1.vsix)(vscode-xml-maven with https://github.com/angelozerr/vscode-xml-maven/blob/master/vscode-xml-maven-0.0.1.vsix)
 * launch vscode-xml with this PR
 * open a pom.xml file and open the completion inside scope element, you should have that:

MavenExtensionDemo

src/extension.ts
```diff
@@ -113,6 +114,12 @@ export function activate(context: ExtensionContext) {
         return text;
       };
 
+      /*if (extensions.onDidChange) {// Theia doesn't support this API yet
```
fbricon: if Theia doesn't support it, then extensions.onDidChange should be false, so you wouldn't go into the block. So why is this block commented?

angelozerr: I commented that because I had some compilation problem (because vscode @types was not used). Now we use @types it works better, I uncomment this code.

package.json
```diff
@@ -28,10 +28,12 @@
...
     "test-compile": "tsc -p ./",
-    "postinstall": "node ./node_modules/vscode/bin/install",
-    "test": "node ./node_modules/vscode/bin/test",
-    "build-server": "./node_modules/.bin/gulp build_server",
-    "tslint": "./node_modules/.bin/gulp tslint"
+    "pretest": "npm run test-compile",
+    "test": "node ./out/test/runTest.js",
+    "tslint": "./node_modules/.bin/tslint -p .",
+    "build": "./node_modules/.bin/gulp build",
+    "build-server": "./node_modules/.bin/gulp buildServer",
```
xorye: `buildServer` should be `build_server` because the gulp task is called `build_server`: https://github.com/angelozerr/vscode-xml/blob/5b1969aa5a1e5c73a08a00f45ffb68d358cfeb43/gulpfile.js#L13

package.json
```diff
@@ -28,10 +28,12 @@
...
+    "build-ext": "./node_modules/.bin/gulp buildExtension"
```
xorye: please remove this one, there is no `buildExtension` task in the [gulpfile](https://github.com/angelozerr/vscode-xml/blob/5b1969aa5a1e5c73a08a00f45ffb68d358cfeb43/gulpfile.js#L13)

package.json
xorye: alternatively, if you wish, you can rename the  `build_server` task to `buildServer` in the gulpfile

xorye:
the completion is working for me

package.json
```diff
@@ -44,25 +46,31 @@
...
-    "@types/mocha": "^5.2.5",
-    "@types/node": "^6.0.40",
-    "gulp": "^4.0.0",
+    "@types/fs-extra": "^7.0.0",
+    "@types/md5": "^2.1.33",
+    "@types/mocha": "^5.2.6",
+    "@types/node": "^10.14.16",
+    "@types/vscode": "^1.37.0",
+    "@types/which": "^1.3.1",
+    "@types/xml2js": "^0.4.4",
```
xorye: We don't need the following dependencies:   
"@types/fs-extra": "^7.0.0",
"@types/md5": "^2.1.33",
"@types/which": "^1.3.1",
"@types/xml2js": "^0.4.4",

package.json
angelozerr: Good catch. It was a copy/paste mistake. It should be fixed now.

package.json
angelozerr: Good catch. It was a copy/paste mistake. It should be fixed now. I have just one question, how have you seen the trouble (have you some error lint you could have to see unused dependencies?)

package.json
angelozerr: done

angelozerr:
> the completion is working for me

Are you testing in Linux OS? If your test in Linux OS, it's very cool because it means classpath build is working on any OS (I'm on Windows OS). See https://github.com/redhat-developer/vscode-xml/pull/193/files#diff-d106f0e1914f2f6cbf79404a7f0c42b4R43

@xorye please tell me if you need some other changes, otherwise please merge it, thanks!

xorye:
Yes, I am testing on Linux OS, and `pathSeparator` is `:` for me, so that's good.

xorye:
https://github.com/redhat-developer/vscode-xml/blob/3d321fe82e8060255d67e2a22bc496d9656b44de/package.json#L19

Could you change this to "^1.37.0"?
Since we are using @types/vscode 1.37.0, we need that to change.

angelozerr:
> Could you change this to "^1.37.0", so that its "vscode": "^1.37.0"?

@xorye fixed.

Issue #149, opened 2019-06-04
Gracefully handle cancelled  'xml/closeTag' requests


fbricon:
With https://github.com/angelozerr/lsp4xml/pull/412 enabled, rapidly typing closing tags in a large xml document leads to vscode logging errors:
```
rejected promise not handled within 1 second: Error: The request (id: 19, method: 'xml/closeTag') has been cancelled
extensionHostProcess.js:749
stack trace: Error: The request (id: 19, method: 'xml/closeTag') has been cancelled
	at handleResponse (/Users/fbricon/Dev/projects/vscode-xml/dist/extension.js:6061:48)
	at processMessageQueue (/Users/fbricon/Dev/projects/vscode-xml/dist/extension.js:5888:17)
	at Immediate.module.exports../node_modules/vscode-jsonrpc/lib/main.js.setImmediate (/Users/fbricon/Dev/projects/vscode-xml/dist/extension.js:5872:13)
	at runCallback (timers.js:696:18)
	at tryOnImmediate (timers.js:667:5)
	at processImmediate (timers.js:649:5)
```

Cancellations should be handled more gracefully

PR #204, opened 2019-12-09
Allows editing of start/end tag simultaneously

### LINKED ISSUE #130: Auto Rename Tag
I just realized, with the help of [@angelozerr](https://github.com/angelozerr) at [lsp4xml/#362](https://github.com/angelozerr/lsp4xml/issues/362), that synced renaming and deletion of two tags, that belong together, may best be done via some simple text manipulation on the client side. I now have installed a little extension into VSCode, that does just that (minus paired delete, sadly), but it may well fit also into your project, so to keep things together.

NikolasKomonen:
Under the preference xml.autoSelectingMatchingTags

Fixes #130

Signed-off-by: Nikolas Komonen <nikolaskomonen@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,6 +192,12 @@
           "description": "Enable/disable autoclosing of XML tags. \n\nIMPORTANT: Turn off editor.autoClosingTags for this to work",
           "scope": "window"
         },
+        "xml.matchingTagEditing": {
+          "type": "boolean",
+          "scope": "window",
+          "default": true,
+          "description": "Adds an additional cursor on the matching tag, allows for start/end tag editing."
+        },
         "xml.codeLens.enabled": {
           "type": "boolean",
           "default": false,
@@ -293,6 +299,14 @@
       {
         "fileMatch": "package.json",
         "url": "./schemas/package.schema.json"
+      }
+    ],
+    "keybindings":[
+      {
+        "command": "xml.toggleMatchingTagEdit",
+        "key": "ctrl+shift+f2",
+        "mac": "cmd+shift+f2",
+        "when": "editorFocus"
       }
     ]
   }
--- a/src/mirrorCursor.ts
+++ b/src/mirrorCursor.ts
@@ -1,0 +1,254 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import {
+	window,
+	workspace,
+	Disposable,
+	TextDocument,
+	Position,
+	TextEditorSelectionChangeEvent,
+	Selection,
+	Range,
+	WorkspaceEdit
+} from 'vscode';
+
+export function activateMirrorCursor(
+	matchingTagPositionProvider: (document: TextDocument, position: Position) => Thenable<Position | null>,
+	supportedLanguages: { [id: string]: boolean },
+	configName: string
+): Disposable {
+	let disposables: Disposable[] = [];
+
+	window.onDidChangeTextEditorSelection(event => onDidChangeTextEditorSelection(event), null, disposables);
+	let previousState = workspace.getConfiguration().get<boolean>(configName);
+	let wasNotified = false;
+	let isEnabled = false;
+	updateEnabledState();
+
+	workspace.onDidChangeConfiguration(updateEnabledState, null, disposables);
+
+	function updateEnabledState() {
+		updateStateSetting();
+		promptUpdateMessage();
+	}
+
+	function updateStateSetting() {
+		isEnabled = false;
+		let editor = window.activeTextEditor;
+		if (!editor) {
+			return;
+		}
+		let document = editor.document;
+		if (!supportedLanguages[document.languageId]) {
+			return;
+		}
+		if (!workspace.getConfiguration(undefined, document.uri).get<boolean>(configName)) {
+			return;
+		}
+		isEnabled = true;
+	}
+
+	function promptUpdateMessage() {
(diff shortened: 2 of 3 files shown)


src/matchingTag.ts
```diff
@@ -0,0 +1,147 @@
+/*---------------------------------------------------------------------------------------------
```
angelozerr: @NikolasKomonen it seems that you are using an old code (which I think that there are some bugs). IMHO you should use https://github.com/microsoft/vscode/blob/master/extensions/html-language-features/client/src/mirrorCursor.ts

fbricon:
Can't get it to work. Output filled with errors like:
```
rejected promise not handled within 1 second: Error: Unsupported request method: xml/matchingTagPosition
extensionHostProcess.js:820
stack trace: Error: Unsupported request method: xml/matchingTagPosition
	at handleResponse (/Users/fbricon/Dev/projects/vscode-xml/dist/extension.js:6179:48)
	at processMessageQueue (/Users/fbricon/Dev/projects/vscode-xml/dist/extension.js:6006:17)
	at Immediate.<anonymous> (/Users/fbricon/Dev/projects/vscode-xml/dist/extension.js:5990:13)
	at processImmediate (internal/timers.js:439:21)
```

fbricon:
When typing a space at the end of a tag, to add an attribute for instance, there's often an error logged because of cancelled request. Logs fill up pretty quickly so this should be handled more gracefully:
Dec-10-2019 16-16-23

NikolasKomonen:
> When typing a space at the end of a tag, to add an attribute for instance, there's often an error logged because of cancelled request. Logs fill up pretty quickly so this should be handled more gracefully:

@fbricon I can't recreate this, but maybe it is using the bugged version of the code. I pushed the new one here, just pull the client again and hopefully this disappears.

fbricon:
Works beautifully now, can't reproduce the issue from earlier. 

That's a great feature, nice job @NikolasKomonen!

fbricon:
Spoke too soon, I can still reproduce. add space at the end of a tag, delete, rinse, repeat until the exception shows up

fbricon:
definitely reminds me of #149. We prolly need to make the server side more robust wrt cancelled commands

fbricon:
I think the key here is "rejected promise not handled within 1 second". Server is too slow to respond, then client cancels the request.

fbricon:
See https://stackoverflow.com/a/50124430/753170 for a potential solution

NikolasKomonen:
> See https://stackoverflow.com/a/50124430/753170 for a potential solution

Yup, this worked for me.

See: https://github.com/redhat-developer/vscode-xml/pull/204/files#diff-2a8a3831675ce7f9a5318a33b0485d7dR86

fbricon:
No this doesn't work for me

fbricon:
Now that've been playing with it, I think we need a new command to toggle the preference, so that, when bound to a keystroke, you can enter/exit quick tag edit mode

Also, the readme needs to be updated with the preference key (and documenting that new toggle command would help too)

NikolasKomonen:
>  I think we need a new command to toggle the preference

Right now you can press Esc to stop the edit mode. Is that good enough?

Also, I have the README updated here in the PR here: https://github.com/redhat-developer/vscode-xml/pull/206/files#diff-04c6e90faac2675aa89e2176d2eec7d8R82

fbricon:
Ctrl+Shift+F2 on linux/windows, Cmd+Shift+F2 on mac, to toggle matching tag editing

package.json
```diff
@@ -192,6 +192,12 @@
           "description": "Enable/disable autoclosing of XML tags. \n\nIMPORTANT: Turn off editor.autoClosingTags for this to work",
           "scope": "window"
         },
+        "xml.autoSelectingMatchingTags": {
```
fbricon: xml.matchingTagEditing

src/mirrorCursor.ts
```diff
@@ -0,0 +1,233 @@
...
+							const newCursor = new Selection(
+								matchingTagPosition.line,
+								matchingTagPosition.character,
+								matchingTagPosition.line,
+								matchingTagPosition.character
+							);
+							window.activeTextEditor.selections = [...window.activeTextEditor.selections, newCursor];
+						}
+					}
+				}).then(err => { return; }); // mutes "rejected promise not handled within 1 second"
```
fbricon:
```typescript
.then(undefined,  err => {
	const msg = err.message ;
	// mutes "rejected promise not handled within 1 second"
	if (msg && !msg.endsWith('has been cancelled')){
		console.log(err);
	}
	return; 
});
```

NikolasKomonen:
@fbricon updated everything. The console error should be gone, cmd+shift+f2 should toggle the preference on your machine and a message should appear.

fbricon:
please update the readme

NikolasKomonen:
> please update the readme

@fbricon the readme is updated in this PR: https://github.com/redhat-developer/vscode-xml/pull/206

fbricon:
on configuration change, toggling the new pref should add/remove the mirror cursor immediately

package.json
```diff
@@ -192,6 +192,12 @@
           "description": "Enable/disable autoclosing of XML tags. \n\nIMPORTANT: Turn off editor.autoClosingTags for this to work",
           "scope": "window"
         },
+        "xml.matchingTagEditing": {
```
angelozerr: what about using the same name than vscode `mirrorCursorOnMatchingTag` ?

xorye:
> on configuration change, toggling the new pref should add/remove the mirror cursor immediately

I implemented that here: https://github.com/xorye/vscode-xml/tree/matchingTagClient

fbricon:
superseded by #207

PR #207, opened 2019-12-13
Allows editing of start/end tag simultaneously

### LINKED ISSUE #130: Auto Rename Tag

xorye:
Fixes #130, branched off the branch from https://github.com/redhat-developer/vscode-xml/pull/204

This PR updates the cursor(s) when toggling the config

Signed-off-by: Nikolas Komonen <nikolaskomonen@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,6 +192,12 @@
           "description": "Enable/disable autoclosing of XML tags. \n\nIMPORTANT: Turn off editor.autoClosingTags for this to work",
           "scope": "window"
         },
+        "xml.mirrorCursorOnMatchingTag": {
+          "type": "boolean",
+          "scope": "resource",
+          "default": false,
+          "description": "Adds an additional cursor on the matching tag, allows for start/end tag editing."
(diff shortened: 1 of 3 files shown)


fbricon:
when no folder/workspace is opened, if you edit an xml file, the toggle command doesn't work
If a pref key exists in the workspace, then toggling mirror mode should change it, else, set the preference at the user level

xorye:
@fbricon updated

fbricon:
says "Toggled the `xml.matchingTagEditing` preference in the Workspace settings.", even though there's no workspace

PR #213, opened 2019-12-18
Mirror cursors

xorye:
Here is some work for #130 

Everything that works for the html mirror tags, should work with this PR.

Toggle is available with ctrl+shift+f2.

How it works:
There is a listener that listens for changes in the `xml.mirrorCursorOnMatchingTag` config. This listener is active as long as vscode-xml is activated.

If `xml.mirrorCursorOnMatchingTag` is turned on, a MirrorCursors object is created. This object creates more listeners.
When `xml.mirrorCursorOnMatchingTag` is turned off, the MirrorCursors object (if exists) is disposed, along with the listeners it created.

I must clean the code and play around with it more.
Please feel free to test things out as it is.

### DIFF
--- a/src/mirrorCursor.ts
+++ b/src/mirrorCursor.ts
@@ -1,0 +1,457 @@
+import {
+  window,
+  workspace,
+  ConfigurationChangeEvent,
+  Disposable,
+  TextDocument,
+  TextEditorSelectionChangeEvent,
+  Position,
+  TextEditor,
+  Selection,
+  Range,
+  WorkspaceEdit
+} from 'vscode';
+
+let mirrorCursorsUtils: MirrorCursorsUtils;
+let mirrorCursors: MirrorCursors | undefined;
+
+export interface PositionInterface {
+  line: number;
+  character: number;
+}
+
+export function setupMirrorCursor(matchingTagPositionProvider: (document: TextDocument, position: Position) => Thenable<PositionInterface | null>, supportedLanguages: string[]): Disposable {
+  mirrorCursorsUtils = new MirrorCursorsUtils(matchingTagPositionProvider, supportedLanguages);
+
+  if (mirrorCursorsUtils.mirrorConfigOn()) {
+    mirrorCursors = new MirrorCursors();
+  }
+
+  return workspace.onDidChangeConfiguration(
+    (event: ConfigurationChangeEvent) => {
+      if (!event.affectsConfiguration('xml.mirrorCursorOnMatchingTag')) {
+        return;
+      }
+
+      if (mirrorCursorsUtils.mirrorConfigOn()) {
+        if (mirrorCursors) {
+          throw 'Mirror cursors turned on but mirror cursors already on. This should never happen';
+        }
+        mirrorCursors = new MirrorCursors();
+      } else {
+        if (!mirrorCursors) {
+          throw 'Mirror cursors turned off but mirror cursors were already off. This should never happen';
+        }
+        mirrorCursors.cleanup();
+        mirrorCursors = undefined;
+      }
+    }
+  );
+}
+
+class MirrorCursorsUtils {
+
+  private matchingTagPositionProvider: (document: TextDocument, position: Position) => Thenable<PositionInterface | null>;
+  private supportedLanguages;
+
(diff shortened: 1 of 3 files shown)


angelozerr:
@xorye IMHO I think we should keep the original mirrorCursor.ts from vscode and update it by adding comments (// updated for vscode-xml) because:

 * it will be more easy to review your PR (just comparing original vscode mirrorCursor.ts with your mirrorCursor.ts)
 * when vscode will fix something in mirrorCursor.ts, it will be easy to fix it in vscode-xml if we have the same code.

@fbricon what do you think about that?

src/mirrorCursor.ts
```diff
@@ -0,0 +1,461 @@
...
+  }
+
+  private supportedFileToSupported(): boolean {
+    return this.currentFileIsSupported && !!this.cursorUpdater;
+  }
+
+  public cleanup(): void {
+    if (this.cursorUpdater) this.cursorUpdater.cleanup(true);
+    this.dispose();
+    window.showInformationMessage('Mirror cursor destroyed!');
```
fbricon: please remove this

src/mirrorCursor.ts
```diff
@@ -0,0 +1,461 @@
...
+
+    return document.getText(new Range(position, document.positionAt(offset + 1)));
+  }
+
+  public cleanup(forceOneCursor = false): void {
+    if (forceOneCursor && window.activeTextEditor && window.activeTextEditor.selections.length > 0) {
+      window.activeTextEditor.selections = [window.activeTextEditor.selections[0]];
+    }
+    this.dispose();
+    window.showInformationMessage('Cursor updater destroyed!');
```
fbricon: please remove this

src/mirrorCursor.ts
```diff
@@ -0,0 +1,461 @@
...
+    }
+    this.prevCursors = [];
+    this.currCursors = [];
+    this.disposables = [];
+    this.isMirrorCursorsOn = false;
+
+    this.disposables.push(
+      window.onDidChangeTextEditorSelection(this.updateCursors.bind(this))
+    );
+    window.showInformationMessage('Cursor updater instantiated!');
```
fbricon: please remove this

src/mirrorCursor.ts
```diff
@@ -0,0 +1,461 @@
...
+    this.currentFileIsSupported = mirrorCursorsUtils.isEditorSupportedLanguage(
+      window.activeTextEditor
+    );
+    this.disposables = [];
+    this.setupSupportedFileListener();
+
+    if (this.currentFileIsSupported) {
+      this.cursorUpdater = new CursorUpdater();
+    }
+    window.showInformationMessage('Mirror cursor instantiated!');
```
fbricon: please remove this

fbricon:
new version seems way safer

xorye:
fixed

fbricon:
We still have that completion issue, that the html server doesn't have:
Jan-28-2020 17-11-24

xorye:
The reason why it works for the HTML language server is that for their completion requests, the new text that is being inserted is only the tag name (ie, `div`).

LSP4XML provides the brackets, start tag and end tag on completion, (ie `<div></div>`). One solution would be to have LSP4XML do the same thing as the HTML language server and just send back the tag names, but I'm not sure if that is feasible in the context of XML?

angelozerr:
As @xorye LSP4XML generate the start and end tag when completion is applied (like Eclipse WTP) although HTML Language server generates only the start tag.

An idea of fix is to check if mirror cursor is activated and send this settings to the LSP4XML. In this case, the end tag should not be generated and we will have the same context than HTML language server.

fbricon:
I think you can check whether the tag is closed or not, whether mirror cursors is enabled or not

angelozerr:
> I think you can check whether the tag is closed or not

Ideally we should do that, but I don't know if it's possible to do that, need more investigation.

fbricon:
See related https://github.com/microsoft/vscode/issues/88424

xorye:
Closing in favour of https://github.com/redhat-developer/vscode-xml/pull/304

Issue #219, opened 2020-01-12
Formatter expand/collapse/ignore empty XML tags


alanlivio:
is It possible to configure the vscode-XML formatter to collapse empty XML tags?
Exemple. from:

```xml
<example></example>
```
to:

```xml
<example/>
```

It is done by other vscode extension but individually:
  https://marketplace.visualstudio.com/items?itemName=wongjiahau.collapsed-expand-empty-tag

angelozerr:
Indeed we should provide this format settings but we should take care if XML is bound to a XML Schema /DTD which defines non empty element.

angelozerr:
@alanlivio I'm working on this issue, see the current status:

![FormatEmptyElements](https://user-images.githubusercontent.com/1932211/79634954-0f38fc80-816e-11ea-94df-c3054048033d.gif)

angelozerr:
It seems https://github.com/wongjiahau/VSCode-CollapseExpandEmptyTag#known-issues have some limitation, vscode-xml will have not this limitation:

![FormatEmptyElements2](https://user-images.githubusercontent.com/1932211/79635185-bec29e80-816f-11ea-8aa3-46024442b784.gif)

fbricon:
New `xml.format.emptyElements` preference has been added with expand/collapse/ignore options

alanlivio:
Thanks @angelozerr and @fbricon .
Best regards.

PR #241, opened 2020-04-18
Formatter expand/collapse/ignore empty XML tags

angelozerr:
Formatter expand/collapse/ignore empty XML tags

See https://github.com/redhat-developer/vscode-xml/issues/219

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ The following settings are supported:
 * `xml.catalogs` : Register XML catalog files.
 * `xml.logs.client` : Enable/disable logging to the Output view.
 * `xml.fileAssociations` : Associate XML Schemas to XML file patterns.
+* `xml.format.emptyElements` : Expand/collapse empty elements during formatting. Defaults to `ignore`.
 * `xml.format.splitAttributes` : Set to `true` to split node attributes onto multiple lines during formatting. Defaults to `false`.
 * `xml.format.joinCDATALines` : Set to `true` to join lines in CDATA content during formatting. Defaults to `false`.
 * `xml.format.joinContentLines` : Set to `true` to join lines in node content during formatting. Defaults to `false`.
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -106,6 +106,17 @@
           "default": true,
           "description": "Should the server log to client output"
         },
+        "xml.format.formatEmptyElements": {
```
fbricon, line 109: xml.format.emptyElements

there's already "format" in the key

README.md
```diff
@@ -53,6 +53,7 @@ The following settings are supported:
 * `xml.catalogs` : Register XML catalog files.
 * `xml.logs.client` : Enable/disable logging to the Output view.
 * `xml.fileAssociations` : Associate XML Schemas to XML file patterns.
+* `xml.format.formatEmptyElements` : Expand/Collapse empty elements during formatting. Defaults to `none`.
```
fbricon, line 56: `xml.format.emptyElements`

package.json
angelozerr: fixed in this PR and fixed on LemMinx side

README.md
angelozerr: fixed

xorye:
I found a minor edge case when I set xml.format.emptyElements to collapse:

range formatting

After range formatting, I think the result should be:
```
<a>
    <b>
</b>
</a>
```
as only the starting tag was to be formatted

fbricon:
@angelozerr please see if this is fixable. maybe don't do the collapsing/expansion if the  selection is not self-contained, i.e it's malformed on its own.

angelozerr:
I close this PR for https://github.com/redhat-developer/vscode-xml/pull/242 (I had some trouble with Eclipse Egit NPE bug)

angelozerr:
> I found a minor edge case when I set xml.format.emptyElements to collapse:

Good catch!

> @angelozerr please see if this is fixable. maybe don't do the collapsing/expansion if the selection is not self-contained, i.e it's malformed on its own.

I fixed this issue, if element is not closed (when range doesn't cover the whole element, the collapse/expand is ignored for this element

PR #266, opened 2020-05-28
Fix debug mode detection

fbricon:
It was no longer possible to start a remote Java debugger on lemminx,
because the debug flags used to start the hosted vscode extension have changed.


xorye:
Works in both the current VS Code release and the latest Insiders (which uses `--inspect`)

PR #278, opened 2020-06-15
Use JDK default GC

brunoborges:


fbricon:
@brunoborges any specific reason for this change?

brunoborges:
It won't work OOTB with a JDK that does not have G1GC, requiring user to tweak the config.

fbricon:
I'm curious to know which JDKs don't have G1GC today. OracleJDK, OpenJDK, OpenJ9 have it, even with Java 8

brunoborges:
I'm curious to understand then, why the core Java Language Support extension doesn't come with G1GC default setting. :-)

But to your point, a build of OpenJDK doesn't *have* to come with G1GC. Oracle's builds of OpenJDK don't come with Shenandoah for example.

brunoborges:
@fbricon any other take on this?

fbricon:
@rgrunber can you check this one please?

rgrunber:
In redhat-developer/vscode-java#1262 we moved from G1 to ParalellGC to reduce the peak heap size. The goal was to not run out of heap in a container environment where that initial project import/build is what causes problems.

I could test this out and see if there's some downside, but I would assume vscode-xml isn't in the same position of needing such optimizations.

rgrunber (approved):
I'll merge this in, since as I mentioned, we're not in a position to need performance improvements on the LemMinx side, so we might as well ensure we support as many JVMs as we can with default settings.

PR #292, opened 2020-07-16
Proposal: Add an API to expose setting configs

a2975667:
Hello Redhat-developer team,
I'm an intern with Salesforce working to support Salesforce developers on XML files. We thought we could make use of the Redhat XML plugin to accomplish this function. We utilize the XML catalog settings and the XML fileAssociation settings to achieve this. However, we do not want to write directly into the developer's settings.json (even on a workspace level) and hope to accomplish feeding the extension of these settings by calling the extension API.

Here we'd like to make a PR and expose these two APIs that Salesforce extensions could call 
`setXMLCatalog()` and `setXMLFileAssociation()` by passing in a JSON from inside Salesforce's extension.
One challenge we had here is because, in the activation function, the `getXMLSettings` will always source directly from the `settings.json` file. This behavior means that every time `getXMLSettings` gets called, our registered settings will be lost. Is there any advice on how to resolve this, or is there another design that the RedHat-develop team suggests?

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,20 +26,116 @@ export interface ScopeInfo {
   configurationTarget: boolean;
 }
 
+/**
+ * Interface for the FileAssociation shape.
+ * @param systemId - The path to a valid XSD.
+ * @param pattern - The file pattern associated with the XSD.
+ * 
+ * @returns None
+ */
+export interface XMLFileAssociation {
+  systemId: string,
+  pattern: string;
+}
+
+/**
+ * Interface for APIs exposed from the extension.
+ * 
+ * @remarks
+ * A sample code to use these APIs are as following:
+ * const ext = await vscode.extensions.getExtension('redhat.vscode-xml').activate();
+ * ext.addXMLCatalogs(...);
+ * ext.removeXMLCatalogs(...);
+ * ext.addXMLFileAssociations(...);
+ * ext.removeXMLFileAssociations(...);
+ */
+export interface XMLExtensionApi {
+  /**
+   * Adds XML Catalogs in addition to the catalogs defined in the settings.json file.
+   * 
+   * @remarks
+   * An example is to call this API:
+   * ```ts
+   * addXMLCatalogs(['path/to/catalog.xml', 'path/to/anotherCatalog.xml'])
+   * ``` 
+   * @param catalogs - A list of path to XML catalogs
+   * @returns None
+   */
+  addXMLCatalogs(catalogs: string[]): void;
+  /**
+   * Removes XML Catalogs from the extension.
+   * 
+   * @remarks
+   * An example is to call this API:
+   * ```ts
+   * removeXMLCatalogs(['path/to/catalog.xml', 'path/to/anotherCatalog.xml'])
+   * ``` 
+   * @param catalogs - A list of path to XML catalogs
+   * @returns None
+   */
+  removeXMLCatalogs(catalogs: string[]): void;
+  /**
+   * Adds XML File Associations in addition to the catalogs defined in the settings.json file.
+   * 
+   * @remarks
+   * An example is to call this API:
+   * ```ts
+   * addXMLFileAssociations([{
+   *    "systemId": "path/to/file.xsd", 
+   *    "pattern": "file1.xml"
+   *  },{
+   *    "systemId": "http://www.w3.org/2001/XMLSchema.xsd",
+   *    "pattern": "file2.xml"
+   *  }])
+   * ``` 
+   * @param fileAssociations - A list of file association 
+   * @returns None
+   */
+  addXMLFileAssociations(fileAssociations: XMLFileAssociation[]): void;
+  /**
+   * Removes XML File Associations from the extension.
+   * 
+   * @remarks
+   * An example is to call this API:
+   * ```ts
+   * removeXMLFileAssociations([{
+   *    "systemId": "path/to/file.xsd", 
+   *    "pattern": "file1.xml"
+   *  },{
+   *    "systemId": "http://www.w3.org/2001/XMLSchema.xsd",
+   *    "pattern": "file2.xml"
+   *  }])
+   * ``` 
+   * @param fileAssociations - A list of file association 
+   * @returns None
+   */
+  removeXMLFileAssociations(fileAssociations: XMLFileAssociation[]): void;
+  /**
+   * Returns the status of the Language Client
+   * 
+   * @remarks
+   * An example is to call this API:
+   * ```ts
+   * isReady()
+   * ``` 
+   * @returns Boolean
+   */
+  isReady(): boolean;
+}
+
 namespace TagCloseRequest {
   export const type: RequestType<TextDocumentPositionParams, AutoCloseResult, any, any> = new RequestType('xml/closeTag');
 }
 
-
 namespace SymbolsLimitExceededNotification {
-  export const type: NotificationType<{commandId: string, message: string}, any> = new NotificationType('xml/symbolsLimitExceeded');
+  export const type: NotificationType<{ commandId: string, message: string }, any> = new NotificationType('xml/symbolsLimitExceeded');
 }
 
 interface ActionableMessage {
-	severity: MessageType;
-	message: string;
-	data?: any;
-	commands?: Command[];
+  severity: MessageType;
+  message: string;
+  data?: any;
+  commands?: Command[];
 }
 
 namespace ActionableNotification {
@@ -48,6 +144,11 @@ namespace ActionableNotification {
 
 export function activate(context: ExtensionContext) {
   let storagePath = context.storagePath;
+  const externalXmlSettings = {
+    "xmlCatalogs": [],
+    "xmlFileAssociations": []
+  };
+
   if (!storagePath) {
     storagePath = os.homedir() + "/.lemminx";
   }
@@ -63,6 +164,7 @@ export function activate(context: ExtensionContext) {
     // rethrow to disrupt the chain.
     throw error;
   }).then(requirements => {
+
     let clientOptions: LanguageClientOptions = {
       // Register the server for xml and xsl
       documentSelector: [
@@ -105,6 +207,7 @@ export function activate(context: ExtensionContext) {
     let languageClient = new LanguageClient('xml', 'XML Support', serverOptions, clientOptions);
     let toDispose = context.subscriptions;
     let disposable = languageClient.start();
+    let clientIsReady = false;
     toDispose.push(disposable);
     languageClient.onReady().then(() => {
       //Detect JDK configuration changes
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -63,6 +63,7 @@ export function activate(context: ExtensionContext) {
     // rethrow to disrupt the chain.
     throw error;
   }).then(requirements => {
+    let XMLsettings = getXMLSettings(requirements.java_home);
```
ntotten, line 66: casing

src/extension.ts
```diff
@@ -146,6 +147,34 @@ export function activate(context: ExtensionContext) {
     });
     languages.setLanguageConfiguration('xml', getIndentationRules());
     languages.setLanguageConfiguration('xsl', getIndentationRules());
+
+    const api = {
+      //setXMLCatalog
+      setXMLCatalog: function (externalSettings) {
```
ntotten, line 153: Should it be `setXmlCatalog`?

src/extension.ts
```diff
@@ -146,6 +147,34 @@ export function activate(context: ExtensionContext) {
...
+        externalSettings['xml']['catalogs'].forEach(element => {
+          if (!XMLsettings['xml']['catalogs'].includes(externalSettings['xml']['catalogs'])) {
+            XMLsettings['xml']['catalogs'].push(element);
+          }
+        });
+        languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: XMLsettings });
+        onConfigurationChange();
+      },
+      //set XMLFileAssoication
+      setXMLFileAssociation: function (externalSettings) {
```
ntotten, line 164: Same, should it be `setXmlFileAssociations`?

ntotten:
The use case you are mentioning is exactly what we are trying to solve, code completion for all of our XML metadata - of which there are many.

We did look into building something with LemMinx as you mentioned, but there were really two things that turned us away from that. The first is that (almost) all of our tooling is written in Typescript/Javascript. It's not that we couldn't provide this with Java, we certainly have Java devs, but the barrier and maintenance of that will be significantly higher for our team.

Second, it seemed like writing a whole new LemMinx extension was significantly more effort than just providing references to XSD files. Right now, we have a very similar process for adding schema validation for our JSON files with VS Code's built-in JSON language server. You can see that adding the schemas is really just a simple file pattern reference: https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/package.json#L813-L822 That is really the type of experience we were hoping for with XML as well.

I'm certainly open to discussion, but my feeling here is that this type of API would be very popular with VS Code extension developers - and while I understand it means it isn't cross-IDE compatible, I think that it still adds a lot of value to those using VS Code so the tradeoff is justified.

angelozerr:
> We did look into building something with LemMinx as you mentioned, but there were really two things that turned us away from that. The first is that (almost) all of our tooling is written in Typescript/Javascript. It's not that we couldn't provide this with Java, we certainly have Java devs, but the barrier and maintenance of that will be significantly higher for our team.

Ok I understand what you mean, but do you think you wanted to provide custom completion (when XML Schema is not enough and values are dynamic (like goal of pom.xml)) ? If it is the case, I think Java dev will be required.

>  You can see that adding the schemas is really just a simple file pattern reference: https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-core/package.json#L813-L822 That is really the type of experience we were hoping for with XML as well.

Indeed I love this JSON feature which is so powerfull and simple. Why not having this same feature? I mean instead of declaring XML catalog, file associations in a settings.json, we could declare it in the package.json. When the XML Language server is started, the settings sent to the server will be the result of merge from package.json and user settings. What do you think about this idea? @fbricon what do you think?

The only thing to do on LemMinx side is to distinguish user settings from static settings which cannot be changed (coming from package.json in vscode context).

src/extension.ts
```diff
@@ -146,6 +147,34 @@ export function activate(context: ExtensionContext) {
...
     languages.setLanguageConfiguration('xml', getIndentationRules());
     languages.setLanguageConfiguration('xsl', getIndentationRules());
+
+    const api = {
+      //setXMLCatalog
+      setXMLCatalog: function (externalSettings) {
+
+        externalSettings['xml']['catalogs'].forEach(element => {
+          if (!XMLsettings['xml']['catalogs'].includes(externalSettings['xml']['catalogs'])) {
+            XMLsettings['xml']['catalogs'].push(element);
```
fbricon, line 157: catalogs.push(element)

src/extension.ts
```diff
@@ -146,6 +147,34 @@ export function activate(context: ExtensionContext) {
...
+      },
+      //set XMLFileAssoication
+      setXMLFileAssociation: function (externalSettings) {
+        externalSettings['xml']['fileAssociations'].forEach(element => {
+          if (!XMLsettings['xml']['fileAssociations'].some((element: { systemId: string; }) => {
+            return element.systemId === externalSettings['xml']['fileAssociations']['systemId'];
+          })) {
+            XMLsettings['xml']['fileAssociations'].push(element);
+          }
+        });
```
fbricon, line 171: use consts, similar to comments about addXMLCatalogs code

fbricon:
I think you need to keep track of the contributed settings somewhere, then merge them with the regular settings whenever those regular settings change, so they can be sent to the server with https://github.com/redhat-developer/vscode-xml/blob/8e7079dab61e22eb8ed79216cccc002527145515/src/extension.ts#L98

angelozerr:
@a2975667 @ntotten what about declaring file association and XML catalog in package.json like JSON does?

IMHO I think it should be a better idea to do that for you to be consistent with JSON. 

I would prefer having this feature instead of exposing a new API (if @fbricon is OK for that)

fbricon:
contributing file associations via package.json config is indeed preferable, if possible, since we wouldn't need to expose/maintain a new API. 
The new API would be useful for dynamic registration though. It all depends on @ntotten @a2975667 needs. No need to get ahead of ourselves if we don't need it ;-)

ntotten:
> Ok I understand what you mean, but do you think you wanted to provide custom completion (when XML Schema is not enough and values are dynamic (like goal of pom.xml)) ? If it is the case, I think Java dev will be required.

It's certainly possible in the future but it really isn't something on our radar. I can maybe think of a few scenarios where we could provide richer completion (i.e. things that are specific for a particular instance or something), but almost all of the completions are standard from static XSDs.

> Indeed I love this JSON feature which is so powerfull and simple. Why not having this same feature? I mean instead of declaring XML catalog, file associations in a settings.json, we could declare it in the package.json. When the XML Language server is started, the settings sent to the server will be the result of merge from package.json and user settings. What do you think about this idea? @fbricon what do you think?

I love the idea of supporting both. I think though for us putting everything in the package.json wont be practical since there are going to be hundreds of different schemas. I think we would want a way to register them dynamically. This actually matches what MS has done with the HTML language server where you can contribute web components for code completion. They allow registering in `settings.json`, through the extension contribution point in `package.json` or through the API. https://code.visualstudio.com/blogs/2020/02/24/custom-data-format

angelozerr:
> I love the idea of supporting both. 

Great.

> I think though for us putting everything in the package.json wont be practical since there are going to be hundreds of different schemas. I think we would want a way to register them dynamically. 

Ok thank's for your clarification, I understand more now why you need this dynamic feature.

src/extension.ts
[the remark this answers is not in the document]
angelozerr: Why add? I think it should be better to keep set because it set the XML catalog and update LSP configuration?

src/extension.ts
```diff
@@ -146,6 +147,34 @@ export function activate(context: ExtensionContext) {
     });
     languages.setLanguageConfiguration('xml', getIndentationRules());
     languages.setLanguageConfiguration('xsl', getIndentationRules());
+
+    const api = {
```
angelozerr, line 151: As it's an API, I think we should defines and export an interface like:

```ts
export interface XMLFileAssociation [
  systemId: string,
  patter; string;
}

export interface XMLExtensionApi {
    setXMLCatalogs (catalogs : string[]),
    setXMLFileAssociations (fileAssociations : XMLFileAssociation[])
}
```

src/extension.ts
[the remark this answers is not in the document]
angelozerr: Why add? I think it should be better to keep set because it set the XML file associations and update LSP configuration?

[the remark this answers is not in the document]
fbricon: Because it adds (.push) the settings to the existing user settings

angelozerr:
@a2975667 @ntotten once this PR will be merged, please update the wiki at https://github.com/redhat-developer/vscode-xml/wiki/Extensions#xml-extension-api-typescript

src/extension.ts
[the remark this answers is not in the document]
angelozerr: Ok so it means that we need a remove method?

src/extension.ts
```diff
@@ -26,20 +26,31 @@ export interface ScopeInfo {
   configurationTarget: boolean;
 }
 
+export interface XMLFileAssociation {
```
angelozerr, line 29: As it's an API, add Javadoc

src/extension.ts
```diff
@@ -26,20 +26,31 @@ export interface ScopeInfo {
...
+  systemId: string,
+  pattern: string;
+}
+
+export interface XMLExtensionApi {
```
angelozerr, line 34: As it's an API, add Javadoc to explain this API. I think we should write in the description how to consume this API by given a sample in the comments.

src/extension.ts
```diff
@@ -26,20 +26,31 @@ export interface ScopeInfo {
...
+  addXMLCatalogs(catalogs: string[]): void,
```
angelozerr, line 35: Add Javadoc (description + @param) for each methods

src/extension.ts
```diff
@@ -146,6 +163,57 @@ export function activate(context: ExtensionContext) {
     });
     languages.setLanguageConfiguration('xml', getIndentationRules());
     languages.setLanguageConfiguration('xsl', getIndentationRules());
+
+    const api: XMLExtensionApi = {
+      // add API set catalogs to internal memory
+      addXMLCatalogs: (catalogs: string[]) => {
+        const externalXmlCatalogs = externalXmlSettings['xmlCatalogs'];
```
angelozerr, line 170:
```ts
const externalXmlCatalogs = externalXmlSettings.xmlCatalogs;
```

src/extension.ts
```diff
@@ -48,6 +59,11 @@ namespace ActionableNotification {
 
 export function activate(context: ExtensionContext) {
   let storagePath = context.storagePath;
+  let externalXmlSettings = {
```
angelozerr, line 62: `const externalXmlSettings`

src/extension.ts
```diff
@@ -146,6 +163,57 @@ export function activate(context: ExtensionContext) {
...
+            externalXmlCatalogs.push(element);
+          }
+        });
+        languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getXMLSettings(requirements.java_home) });
+        onConfigurationChange();
+      },
+      // remove API set catalogs to internal memory
+      removeXMLCatalogs: (catalogs: string[]) => {
+        catalogs.forEach(element => {
+          const externalXmlCatalogs = externalXmlSettings['xmlCatalogs'];
```
angelozerr, line 182: const externalXmlCatalogs = externalXmlSettings.xmlCatalogs;

src/extension.ts
```diff
@@ -146,6 +163,57 @@ export function activate(context: ExtensionContext) {
...
+          if (externalXmlCatalogs.includes(element)) {
+            const itemIndex = externalXmlSettings['xmlCatalogs'].indexOf(element);
```
angelozerr, line 184: const itemIndex = externalXmlCatalogs.indexOf(element);

src/extension.ts
```diff
@@ -146,6 +163,57 @@ export function activate(context: ExtensionContext) {
...
+            externalXmlCatalogs.splice(itemIndex, 1);
+          }
+        });
+        languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getXMLSettings(requirements.java_home) });
+        onConfigurationChange();
+      },
+      // add API set fileAssociations to internal memory
+      addXMLFileAssociations: (fileAssociations: XMLFileAssociation[]) => {
+        const externalfileAssociations = externalXmlSettings['xmlFileAssociations']
```
angelozerr, line 193: `const externalfileAssociations = externalXmlSettings.xmlFileAssociations;`

src/extension.ts
```diff
@@ -146,6 +163,57 @@ export function activate(context: ExtensionContext) {
...
+          if (!externalfileAssociations.some(fileAssociation => fileAssociation.systemId === element.systemId)) {
+            externalfileAssociations.push(element);
+          }
+        });
+        languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getXMLSettings(requirements.java_home) });
+        onConfigurationChange();
+      },
+      // remove API set fileAssociations to internal memory
+      removeXMLFileAssociations: (fileAssociations: XMLFileAssociation[]) => {
+        const externalfileAssociations = externalXmlSettings['xmlFileAssociations']
```
angelozerr, line 204: const externalfileAssociations = externalXmlSettings.xmlFileAssociations;

src/extension.ts
```diff
@@ -189,6 +257,19 @@ export function activate(context: ExtensionContext) {
     xml['xml']['format']['trimFinalNewlines'] = workspace.getConfiguration('files').get('trimFinalNewlines', true);
     xml['xml']['format']['trimTrailingWhitespace'] = workspace.getConfiguration('files').get('trimTrailingWhitespace', false);
     xml['xml']['format']['insertFinalNewline'] = workspace.getConfiguration('files').get('insertFinalNewline', false);
+
+    //applying externalXmlSettings to the xmlSettings
+    externalXmlSettings["xmlCatalogs"].forEach(catalog => {
```
angelozerr, line 262: externalXmlSettings.xmlCatalogs.forEach

src/extension.ts
```diff
@@ -189,6 +257,19 @@ export function activate(context: ExtensionContext) {
...
+      if (!xml['xml']['catalogs'].includes(catalog)) {
+        xml['xml']['catalogs'].push(catalog);
+      }
+    })
+    externalXmlSettings["xmlFileAssociations"].forEach(element => {
```
angelozerr, line 267: externalXmlSettings.xmlFileAssociations.forEach

src/extension.ts
a2975667: I placed the comments on top of each API below

angelozerr: Thanks but my idea is to write a sample code by getting extension and consume the API, so I think the Javadoc should mention this piece of code: 

```ts
const ext = await vscode.extensions.getExtension('redhat.vscode-xml').activate();
ext.addXMLCatalogs...
```

a2975667: I see! let me edit that.

angelozerr:
@a2975667 please click on `Resolve Conversation` once you have fixed a problem, it will help more for the review, thanks!

Once you have fixed my last comment https://github.com/redhat-developer/vscode-xml/pull/292#discussion_r456958904 and if it works for you (even with your usecase whith lost of settings), I approve this PR.

I let @fbricon to merge it if he think it's good.

a2975667:
Please let me know if there are any other things that I should address. Thanks! :)

a2975667:
@fbricon are there any other things that I miss addressing? Thanks so much again!

fbricon:
Please squash your commits into one and use `git commit -s` to sign it

a2975667:
@fbricon Done. Please verify. Thanks!

a2975667:
@angelozerr @fbricon @ntotten 
Added an additional API to return the status of the language client. 
This addition is helpful because there is, to the best of my knowledge, a way to know if the extension's language server is ready or not. The current `.isActive()` seems to only checks if `activate()` was invoked.
Having this additional API makes sure that the extension is ready and one can safely apply the API calls.
Please comment.

fbricon:
@a2975667 can you please sign-off both your commits? (see https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff)

a2975667:
@fbricon Just did. Sorry for that.

a2975667:
@fbricon Please let me know if there are any other things I need to address. Thanks!

fbricon:
Thanks for your contribution @a2975667 !

PR #293, opened 2020-07-17
Dispose extensions did change listener when deactivated

rzgry:
Noticed this when working on vscode-microprofile :) 

Disposes the `extensions.onDidChange` event listener if vscode-xml is deactivated / uninstalled.


angelozerr:
Thanks @rzgry! 

@xorye could you review it please.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vscode-xml",
   "displayName": "XML",
   "description": "XML Language Support by Red Hat",
-  "version": "0.13.0",
+  "version": "0.14.0",
```
xorye, line 5: Thanks for this change. Could we have it in a separate commit, either in this PR or a separate PR?

rzgry: Will add it in a separate PR

xorye:
thanks a lot @rzgry!

PR #297, opened 2020-07-30
Patch: Remove the need for isReady and refactor onReady return API

a2975667:
This PR removes the need for `isReady()` API and instead returns the APIs directly through the built in `onReady` call.
This patch makes using the API more elegant and clean.

Signed-off-by: Ti-Chung Cheng <t.cheng@salesforce.com>


a2975667:
@fbricon Apologies for opening an additional PR patching the previous merged PR. Please review. 
This change makes calling the APIs a lot more elegant and clean.

angelozerr:
@datho7561  could you review this PR if you have time.

PR #300, opened 2020-08-04
Update dependencies to fix security vulnerabilities

datho7561:

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,22 +52,20 @@
     "minimist": "^1.2.5"
   },
   "devDependencies": {
-    "@types/mocha": "^5.2.6",
     "@types/node": "^10.14.16",
     "@types/vscode": "^1.37.0",
     "gulp": "^4.0.2",
     "gulp-rename": "^1.4.0",
     "gulp-tslint": "^8.1.3",
-    "mocha": "^6.2.1",
+    "minimist": ">=1.2.5",
     "serialize-javascript": ">=2.1.1",
     "ts-loader": "^6.0.1",
     "tslint": "^5.12.1",
     "typescript": "^3.4.1",
     "typescript-tslint-plugin": "^0.3.1",
     "vscode-test": "^1.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -48,31 +47,28 @@
...
   ],
-  "resolutions": {
-    "minimist": "^1.2.5"
-  },
   "devDependencies": {
     "@types/mocha": "^5.2.6",
-    "@types/node": "^10.14.16",
-    "@types/vscode": "^1.37.0",
+    "@types/node": "^10.17.28",
+    "@types/vscode": "^1.47.0",
```
xorye, line 53: We should probably run `npm audit fix` instead of `npm update --dev` since the latter updates dependencies that don't have security vulnerabilities (according to `npm audit`) like this one.

datho7561: I've tried running just `npm audit fix`, and it seems to break the extension. I'll try some different tactics to avoid updating all the packages.

datho7561: I updated the webpack module (which requires elliptic) and switched an import from require.js to an ES6 import. It now appears to work.

package.json
```diff
@@ -58,16 +58,16 @@
     "gulp": "^4.0.2",
     "gulp-rename": "^1.4.0",
     "gulp-tslint": "^8.1.3",
-    "mocha": "^6.2.1",
+    "minimist": ">=1.2.5",
+    "mocha": "^8.1.1",
```
xorye, line 62: This made me realize, we don't even use mocha in vscode-xml. We can remove this and the typescript definitions here: 

https://github.com/redhat-developer/vscode-xml/blob/24daed620f18660836830bd126f721f7ccb94c7b/package.json#L55

fbricon: @datho7561 please remove mocha then

datho7561: Sorry, I forgot I hadn't pushed this change yet.

fbricon (approved):

PR #301, opened 2020-08-04
Bump version to 0.14.0 in package.json

rzgry:
Change from https://github.com/redhat-developer/vscode-xml/pull/293#discussion_r465303555

Ups the version to 0.14.0 in the package.json. It is set to that in package-lock, but package lock version gets overwritten when running npm install


xorye:
thanks a lot @rzgry!

PR #304, opened 2020-08-05
Type rename provider [Uses VS Code proposed API]

xorye:
Follows HTML's implementation: https://github.com/microsoft/vscode/commit/01e01b13f88064374ebc5aa789cb0ee76d03a17c

To test this PR, you need the server side PR  as well https://github.com/eclipse/lemminx/pull/836

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "license": "EPL-1.0",
   "bugs": "https://github.com/redhat-developer/vscode-xml/issues",
   "preview": true,
+  "enableProposedApi": true,
   "keywords": [
     "xml",
     "xsl",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -11,9 +11,9 @@
  */
 
 import { prepareExecutable } from './javaServerStarter';
-import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient, DidChangeConfigurationNotification, RequestType, TextDocumentPositionParams, ReferencesRequest, NotificationType, MessageType } from 'vscode-languageclient';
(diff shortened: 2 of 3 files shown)


xorye:
Currently not working, but you need to use VS Code insiders https://code.visualstudio.com/insiders/ and set `editor.renameOnType` to true.

Demo for HTML: https://github.com/microsoft/vscode/issues/88424#issuecomment-600840619

xorye:
It works now:
onTypeRename

But we still suffer from this issue: https://github.com/redhat-developer/vscode-xml/pull/213#issuecomment-579328288
onTypeRenameCompletion

package.json
```diff
@@ -9,6 +9,7 @@
   "license": "EPL-1.0",
   "bugs": "https://github.com/redhat-developer/vscode-xml/issues",
   "preview": true,
+  "enableProposedApi": true,
```
fbricon, line 12: we can only use it in insiders. The marketplace rejects publishing when enableProposedApi:true

angelozerr:
> But we still suffer from this issue: #213 (comment)

It's very strange because this problem should be fixed. The only problem that I have seen is when you do this usecase with a XML element which doesnt comes from a XSD Schema (but it seems that it's not our usecase).

In the demo, I noticed that when you remove the d character completion opens, inmy case the completion is not opened (I must do a Ctrl+Space to open completion). Is there a settings for that?

angelozerr:
We need to wait for https://github.com/microsoft/vscode/issues/105394

package.json
angelozerr: We need to wait for microsoft/vscode#105394

angelozerr:
Please note the onTypeRanme is now in the LSP 3.16 specification. I have created an issue in LSP4J https://github.com/eclipse/lsp4j/issues/469 If I understand correctly, we will  avoid developping TypeScript code for supporting mirror cursor.

angelozerr:
Today this type rename provider became linked editing range with a LSP specification. This PR is replaced with https://github.com/redhat-developer/vscode-xml/pull/432

PR #305, opened 2020-08-07
Stop XML server when extension is deactivated

### LINKED ISSUE #303: Stop the XMLServerLauncher process when the extension is deactivated
Sometimes when you exit VS Code while vscode-xml was running, there is a Java process called `XMLServerLauncher` that remains.

image

If that process remains, and vscode-xml is started up again, there is an error about the XML language server crashing 5 times in the last 3 minutes.

To fix this problem vscode-xml should call `stop()` on the languageClient in the `deactivate()` method, which is what vscode-java does:
https://github.com/redhat-developer/vscode-java/blob/c7afc32bde70f5afc292966c1311ec4dd1bfe9b5/src/extension.ts#L384-L387

to ensure that the process is terminated when the extension is deactivated.

rzgry:
Fixes #303 

Moves the language client declaration to the top level scope so it can be accessed by the `deactivate` function

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -132,6 +132,8 @@ namespace ActionableNotification {
   export const type = new NotificationType<ActionableMessage, void>('xml/actionableNotification');
 }
 
+let languageClient: LanguageClient;
+
 export function activate(context: ExtensionContext) {
   let storagePath = context.storagePath;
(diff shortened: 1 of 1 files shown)


angelozerr:
Thanks @rzgry !

PR #312, opened 2020-08-26
Fix warnings in extension host

### LINKED ISSUE #98: Warnings in extension development
Not a blocking issue. Just some annoying warning messages in debug console if I have xml extension installed. 

```
[redhat.vscode-xml] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '[xml]', provide the URI of a resource or 'null' for any resource.
[redhat.vscode-xml] Accessing a window scoped configuration for a resource is not expected. To associate 'xml.completion.autoCloseTags' to a resource, define its scope to 'resource' in configuration contributions in 'package.json'.
```

### The first one:
https://github.com/redhat-developer/vscode-xml/blob/32b586ce109b9ac6d1fa2a532d1abb4200a8a26d/src/extension.ts#L128
`let configXML = workspace.getConfiguration(undefined, null)`, `ConfigXML.get("[xml]")` should work.

### The second one:
https://github.com/redhat-developer/vscode-xml/blob/32b586ce109b9ac6d1fa2a532d1abb4200a8a26d/src/extension.ts#L86
https://github.com/redhat-developer/vscode-xml/blob/407cd654cc7d237601d7a926b9eb1773816cdb1a/src/tagClosing.ts#L32
https://github.com/redhat-developer/vscode-xml/blob/32b586ce109b9ac6d1fa2a532d1abb4200a8a26d/package.json#L152-L156

datho7561:
Use method described by Eskibear to access [xml] autoclosing settings.
Do not provide a document when accessing vscode-xml settings, since LemMinX doesn't support multi-root projects right now.

Closes #98

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -51,7 +51,7 @@ export function subscribeJDKChangeConfiguration() {
       oldXMLConfig = newXMLConfig;
       return;
     }
-    
+
     //handle "java.home" change
     if(oldXMLConfig.get("java.home") == null) { // if "xml.java.home" exists, dont even look at "java.home"
       if(params.affectsConfiguration("java")) {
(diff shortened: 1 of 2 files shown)


Eskibear:
Since all the settings here are in `window` scope, you should not pass the document. See: https://github.com/microsoft/vscode/blob/40827f365c376dd6753116326fd10a9f826655b7/src/vs/vscode.d.ts#L9952-L9965

Take `xml.completion.autoCloseTags` for example, `workspace.getConfiguration("")` should work. But I prefer below to match the name `configXML`.
```
configXML = workspace.getConfiguration("xml")
configXML.get("completion.autoCloseTags")  
```

And I guess that would be more efficient as you don't fetch all the other (unrelated) settings into `configXML`. 

And for per-language settings under "[xml]" , looks that you'd better use`workspace.getConfiguration(section, {languageId: "xml"})` according to https://github.com/microsoft/vscode/blob/40827f365c376dd6753116326fd10a9f826655b7/src/vs/vscode.d.ts#L9997-L10003

datho7561:
Thanks for the information!

I have updated the code so that when accessing vscode-xml settings, documents aren't provided.

The [version of the vscode API](https://github.com/redhat-developer/vscode-xml/blob/master/package.json#L22) that the extension is using doesn't yet have https://github.com/microsoft/vscode/blob/40827f365c376dd6753116326fd10a9f826655b7/src/vs/vscode.d.ts#L9997-L10003. It expects the second parameter to be `Uri|null`. Unless we update the API version that vscode-xml uses, there is no clean way to access `[xml].editor.autoClosingBrackets`. I commented the hack that is used to access it (passing `null` as the document).

datho7561:
Since Theia doesn't support this API change yet, passing `null` seems like the best option for now.

datho7561:
Closing this PR, since the implementation is a bit of a hack. The issue can probably be addressed the proper way that Eskibear described, since we now require VS Code `^1.52.0`

PR #334, opened 2020-10-02
Add variables for `xml.fileAssociations`

### LINKED ISSUE #307: Allow support for variable and different file path notations in xml.FileAssociations
A number of things that are normally supported by configuration options in VS code JSON files do not seem to work when configuring xml.FileAssociations. 
Namely:
* Variable Substitution
* Windows-Style file paths. 

As an example, I have a test repository that has a schema in one folder, and an XML file in another folder. 

<details> 
<summary> Project Structure Photograph </summary>

image

</details>

```{
    "xml.fileAssociations": [{
            "systemId": "schema/schema.xsd",
            "pattern": "testFile/*.xml"
        }]
} 
```

This works as expected, XML files in the testFile folder are checked against the schema document. 

But if I instead refer to the files using the `${workspaceFolder}` variable built into VS code like this: 

```{
    "xml.fileAssociations": [{
            "systemId": "${workspaceFolder}/schema/schema.xsd",
            "pattern": "${workspaceFolder}/testFile/*.xml"
        }]
}
```

The files are not being checked against the schema, and I do not get any warnings. 

Furthermore, if I use windows style file path notation

``` {
    "xml.fileAssociations": [{
            "systemId": "schema\\schema.xsd",
            "pattern": "testFile\\*.xml"
        }]
}
```

I once again find that my XML files are not being checked against the schema file. 

Both of these should be supported when configuring settings for the extension. Variable substitution is important for configuring paths in global settings that will change for each workspace, and Windows file path support is necessary to maintain consistency with other options that might be set in the same file.

datho7561:
Adds three variables that can be used in `xml.fileAssociations`:
 * ${workspaceFolder}
 * ${fileDirname}
 * ${fileBasenameNoExtension}

 These variables can be used for both the `pattern` and the `systemId`.

 Closes #307

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,32 +10,18 @@
  *  Microsoft Corporation - Auto Closing Tags
  */
 
-import { prepareExecutable } from './javaServerStarter';
-import {
-  LanguageClientOptions,
-  RevealOutputChannelOn,
-  LanguageClient,
-  DidChangeConfigurationNotification,
-  RequestType,
-  TextDocumentPositionParams,
-  ReferencesRequest,
-  NotificationType,
-  MessageType,
-  ConfigurationRequest,
-  ConfigurationParams,
-  ExecuteCommandParams,
-  CancellationToken,
-  ExecuteCommandRequest, TextDocumentIdentifier
-} from 'vscode-languageclient';
-import * as requirements from './requirements';
-import { languages, IndentAction, workspace, window, commands, ExtensionContext, TextDocument, Position, LanguageConfiguration, Uri, extensions, Command, TextEditor } from "vscode";
-import * as path from 'path';
 import * as os from 'os';
-import { activateTagClosing, AutoCloseResult } from './tagClosing';
+import * as path from 'path';
+import { Command, commands, ExtensionContext, extensions, IndentAction, LanguageConfiguration, languages, Position, TextDocument, TextEditor, Uri, window, workspace } from "vscode";
+import { CancellationToken, ConfigurationParams, ConfigurationRequest, DidChangeConfigurationNotification, ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, MessageType, NotificationType, ReferencesRequest, RequestType, RevealOutputChannelOn, TextDocumentIdentifier, TextDocumentPositionParams } from 'vscode-languageclient';
 import { Commands } from './commands';
-import { getXMLConfiguration, onConfigurationChange, subscribeJDKChangeConfiguration } from './settings';
-import { collectXmlJavaExtensions, onExtensionChange } from './plugin';
+import { prepareExecutable } from './javaServerStarter';
 import { markdownPreviewProvider } from "./markdownPreviewProvider";
+import { collectXmlJavaExtensions, onExtensionChange } from './plugin';
+import * as requirements from './requirements';
+import { getXMLConfiguration, onConfigurationChange, subscribeJDKChangeConfiguration } from './settings';
+import { activateTagClosing, AutoCloseResult } from './tagClosing';
+import { containsVariableReferenceToCurrentFile, getVariableSubstitutedAssociations } from './variableSubstitution';
 
 export interface ScopeInfo {
   scope: "default" | "global" | "workspace" | "folder";
@@ -340,6 +326,14 @@ export function activate(context: ExtensionContext) {
         }
         return result;
       });
+      // When the current document changes, update variable values that refer to the current file if these variables are referenced,
+      // and send the updated settings to the server
+      context.subscriptions.push(window.onDidChangeActiveTextEditor(() => {
+        if (containsVariableReferenceToCurrentFile(getXMLConfiguration().get('fileAssociations') as XMLFileAssociation[])) {
+          languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getXMLSettings(requirements.java_home) });
+          onConfigurationChange();
+        }
+      }));
 
       const api: XMLExtensionApi = {
         // add API set catalogs to internal memory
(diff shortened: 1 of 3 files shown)


datho7561:
This change is client only: the variables only in vscode-xml, not any of the other clients that LemMinX supports.

Still need to:
- [ ] Clean up the code
- [ ] Add documentation in wiki
- [ ] Add documentation to internal documents

fbricon:
please rebase against master

datho7561:
I messed up the rebase, one second...

src/extension.ts
```diff
@@ -420,14 +414,70 @@ export function activate(context: ExtensionContext) {
...
+      xml['xml']['fileAssociations'].map((association: XMLFileAssociation): XMLFileAssociation => {
+
+        const currentFile: TextDocument = window.activeTextEditor.document;
+        const currentFileUri: string = currentFile && currentFile.uri.fsPath;
+        const currentWorkspace: WorkspaceFolder = workspace.getWorkspaceFolder(currentFile && currentFile.uri);
+        const currentWorkspaceUri: string = (currentWorkspace && currentWorkspace.uri.fsPath)
+          || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath);
+
+        if (!currentWorkspaceUri
+          && (association.pattern.indexOf('&{workspaceFolder}') >= 0
```
angelozerr, line 427: I think we should have a global const with variables substitution information (name + referencedCurrentFile) and after that you could use this array to do your logic.

The idea is when we will want to add another variable, we should just update the array.

src/extension.ts
```diff
@@ -420,14 +414,70 @@ export function activate(context: ExtensionContext) {
...
+            || association.systemId.indexOf('&{workspaceFolder}') >= 0)) {
+          return;
+        }
+
+        if (!currentFileUri
+          && (association.pattern.indexOf('&{fileDirname}') >= 0
```
angelozerr, line 433: Use global const with variables substitution information

src/extension.ts
```diff
@@ -420,14 +414,70 @@ export function activate(context: ExtensionContext) {
...
+         * * ${workspaceFolder}
+         * * ${fileDirname}
+         * * ${fileBasenameNoExtension}
+         * substituted into the string
+         *
+         * @param val the value to substitute the variables into
+         * @return the string with values for the variables substituted into the string
+         */
+        const subVars = (val: string): string => {
+          let newVal: string = val.replace(/\$\{workspaceFolder\}/g, currentWorkspaceUri);
```
angelozerr, line 451: Use global const with variables substitution information

src/extension.ts
```diff
@@ -420,14 +414,70 @@ export function activate(context: ExtensionContext) {
...
     return xml;
   }
+
+  /**
+   * Returns true if the the XML settings contain a file association with a variable reference to either ${fileDirname} or ${fileBasenameNoExtension} and false otherwise
+   *
+   * @return true if the the XML settings contain a file association with a variable reference to either ${fileDirname} or ${fileBasenameNoExtension} and false otherwise
+   */
+  function fileAssocReferencesCurrentFile(): boolean {
+    for (const assoc of getXMLConfiguration().get('fileAssociations') as XMLFileAssociation[]) {
```
angelozerr, line 473: Use global const with variables substitution information

src/extension.ts
datho7561: Please let me know if the implementation makes sense.

src/extension.ts
```diff
@@ -420,14 +415,78 @@ export function activate(context: ExtensionContext) {
...
 
     return xml;
   }
+
+  /**
+   * Returns true if the the XML settings contain a file association with a variable reference to either ${fileDirname} or ${fileBasenameNoExtension} and false otherwise
+   *
+   * @return true if the the XML settings contain a file association with a variable reference to either ${fileDirname} or ${fileBasenameNoExtension} and false otherwise
+   */
+  function fileAssocReferencesCurrentFile(): boolean {
```
angelozerr, line 481: * Add  `XMLFileAssociation[] associations` as parameter 
 * Rename `variableReference.ts`  with `variableSubstition.ts`
 * Moce `fileAssocReferencesCurrentFile `into `variableSubstition.ts`

src/variableReference.ts
```diff
@@ -0,0 +1,34 @@
+export class VariableReference {
```
angelozerr, line 1: Rename to variableSubstition.ts

src/extension.ts
```diff
@@ -420,14 +415,78 @@ export function activate(context: ExtensionContext) {
         xml['xml']['catalogs'].push(catalog);
       }
     })
-    externalXmlSettings.xmlFileAssociations.forEach(element => {
-      if (!xml['xml']['fileAssociations'].some(fileAssociation => fileAssociation.systemId === element.systemId)) {
-        xml['xml']['fileAssociations'].push(element);
-      }
-    });
+    const variableSubstitutedAssociations: XMLFileAssociation[] =
```
angelozerr, line 418: Embed the following code in a  function like `getVariableSubstitutedAssociations(XMLFileAssociation[] associations)` that you move into variableSustition.ts

src/extension.ts
angelozerr: It's better, but for `subVars` it's not generic. The idea is to have just an array that you update and that's it. To fix subVars, I suggest that VariableReference defines a method replace or update which update a given string.  Each VariablerReference defines update method with your current replace. After that `subVars`  will just loop for VariablerReference and call the update/replace method.

datho7561: Thanks so much for the review. The code is much easier to read and add onto now. Please let me know if you think its good when you have time.

src/variableSubstitution.ts
```diff
@@ -0,0 +1,154 @@
...
+    return this.replaceRegExp;
+  }
+}
+
+enum VariableSubstitutionKind {
+  Workspace,
+  File
+}
+
+const SETTINGS_VARIABLES: VariableSubstitution[] = [
```
angelozerr, line 68: Very good job! Could you:

 * rename SETTINGS_VARIABLES to VARIABLE_SUBSTITIONS
 * move this const declaration on the top of the file
 * add a comment before the const to explain that of we want to add a new variable substitution, just update this const array.

datho7561: I get a linter error when I try to declare VARIABLE_SUBSTITUTIONS before VariableSubstitution, but I made the other changes.

src/extension.ts
angelozerr: Glad your code pleases you:) It's perfect.

angelozerr (approved):
Great job @datho7561 !

PR #337, opened 2020-10-07
Respond to requests for indentation settings

### LINKED ISSUE #267: EntityNotDeclared codeAction (quick fix) not using proper indentation settings
(Copy-paste from https://github.com/eclipse/lemminx/issues/759)

If `editor.insertSpaces` is set to `false`, the EntityNotDeclared code action does not insert new content with tabs. Instead, the code action inserts new content with spaces. 

Example xml:
```
<?xml version="1.0" standalone="no"?>
<root-element>
&undeclared;
</root-element>
```

Before quick fix:
image

Hover over the error message and invoke the quick fix.

After quick fix:
image

This is because codeAction requests do not receive formatting settings like formatting requests do.
LSP spec: https://microsoft.github.io/language-server-protocol/specification#textDocument_formatting
Related issue: https://github.com/microsoft/language-server-protocol/issues/780

The codeAction is inserting two spaces instead of a single tab, because in the lemminx side, the defaults settings are `insertSpaces: true` and `tabSize: 2`: https://github.com/eclipse/lemminx/blob/84d268957060b53aee211bd75a34b4536e9dab1e/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java#L117-L118

Currently, VS Code is not sending lemminx its values for `editor.insertSpaces` and `editor.tabSize`.

datho7561:
Uses requests to the preferences `xml.format.insertSpaces` and `xml.format.tabSize` to provide the indentation of the current file to the server.

Copied from: https://github.com/redhat-developer/vscode-java/pull/1081/files

Along with eclipse/lemminx#903, closes #267

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -11,9 +11,9 @@
  */
 
 import { prepareExecutable } from './javaServerStarter';
-import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient, DidChangeConfigurationNotification, RequestType, TextDocumentPositionParams, ReferencesRequest, NotificationType, MessageType } from 'vscode-languageclient';
+import { LanguageClientOptions, RevealOutputChannelOn, LanguageClient, DidChangeConfigurationNotification, RequestType, TextDocumentPositionParams, ReferencesRequest, NotificationType, MessageType, ConfigurationRequest, ConfigurationParams } from 'vscode-languageclient';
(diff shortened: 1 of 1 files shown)


angelozerr:
@datho7561 I think we can merge it?

angelozerr (approved):

Issue #220, opened 2020-01-16
Symbol outline / search: Customizable display text


photasm:
It would be nice to have the option to customize how symbol outline / search will be displayed.
This could be done by using the attribute on the displayed element.

```xml
<Element1>
  <Element2 id="id1" />
   <Element2 id="id2" />
   <Element2 id="id3" />
</Element1>
```

In my case my symbol outline looks like this:
```
Element1
--Element2
--Element2
--Element2
````

It would be nice, if it could look sth like this:
```
Element1
--Element2 (id1)
--Element2 (id2)
--Element2 (id3)
````

Thx. :)

**Comment by angelozerr** [comment]
@photasm indeed it should be nice. The main reason why we have not implemented that is that when XML is too big, outline crashes the vscode. 

But since we have added a symbol settings to disable it, I think it's a good idea to add a new settings to add attributes in the outline.

**Comment by photasm** [comment]
Great, thank you in advance.

**Comment by bmix** [comment]
I would like to expand on that. There should be several attributes to configure, which get tested for in sequence.

`@id, @title, @name, @class` are good candidates. If none of them is available, the start of the text node could be shown.

**Comment by angelozerr** [comment]
I did a quick POC for this issue. 

Given the `test.xml` XML file:

```xml
<Element1>
  <Element2 id="id1" />
   <Element2 id="id2" />
   <Element2 id="id3" name="name3"/>
   <Element2>ABCD</Element2>
</Element1>
```
we want to show : 

 * all text node (`ABCD `in this case)
 * only `@id` attribute (not `@name`)

![image](https://user-images.githubusercontent.com/1932211/96035494-63921000-0e63-11eb-8cf2-0dc878229781.png)

To manage this configuration, I introduce a new settings `xml.symbols.filters`:

```json
 "xml.symbols.filters": [
      {
         "pattern": "test.xml",
         "expressions" :[
            {
               "xpath": "//text()",
               "excluded": false
            },
            {
               "xpath": "//@id",
               "excluded": false
            }
         ]
      }
   ]
```

 * pattern means that filters are apply only for file name which match the pattern.
 * expressions are a list of xpath node to include or exclude.

Any feedback are welcome!

PR #344, opened 2020-10-15
Symbol outline / search: Customizable display text

angelozerr:
Symbol outline / search: Customizable display text

See #220


angelozerr:
Replaced with  https://github.com/redhat-developer/vscode-xml/pull/346

PR #358, opened 2020-11-05
fix line breaks, added wiki links

GitMensch:


angelozerr:
@GitMensch thank you for your contribution, but I'm updating README about those links. We decide to give up our wiki and stores documentantion in https://github.com/redhat-developer/vscode-xml/tree/master/docs

The main raison is that vscode-xml provides this documentation inside vscode with the `Open XML Documentation` command.

See https://github.com/redhat-developer/vscode-xml/blob/master/docs/Commands.md

GitMensch:
I see. When will the new README be available? Should I PR line-break only?

angelozerr:
> I see. When will the new README be available? Should I PR line-break only?

Hope I could do that today or tomorrow.

GitMensch:
@angelozerr I see that the new README is up, which is good. But we still have some some links (outline and formatting) linked while some others (which were the initial reason for this PR) aren't in.

Question: Should I redo the PR, this time with the links not to the wiki but the online xml help?

angelozerr:
> Question: Should I redo the PR, this time with the links not to the wiki but the online xml help?

I would like to remove our wiki to use the same documentation for online and embed documentation, so please link to the online xml help. 

We should create the release today, so if you wish to integrate your changes, please create a PR this morning please. Thanks!

GitMensch:
@angelozerr That's fine with me, I'm in the process of doing it.

Question: Do you want to have the "see here" in a separate line (which it partially is, and partially looks like in the README that it is supposed to be, and partially not)? I'd do a cleanup of those in whatever direction you like.
_Personally_ I would otherwise get rid of most of them and instead place the link in the setting name, so instead of

* `xml.symbols.showReferencedGrammars`: Show referenced grammars in the Outline. Default is `true`. See [here](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Symbols.md#xmlsymbolsshowreferencedgrammars) for more information.

do

* [`xml.symbols.showReferencedGrammars`](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Symbols.md#xmlsymbolsshowreferencedgrammars): Show referenced grammars in the Outline. Default is `true`.

Opinions? Wishes?

fbricon:
+1 to unclutter the text and have the links on the pref keys.

GitMensch:
I've found the "con" of the declutter: at least in vscode the links for code entries seem to be only recognizable when hovering over those (get a blue line), other than that they look as they normally do. I still would suggest to do this but would like to get a "go" before doing that...

GitMensch:
New suggestion: I also offer to do the following before the release: move _all_ settings to the onlinedoc, leaving only a reference to this in the README (I find this much easier to read this way, it isn't necessary to adjust the README just because of adding a new one, and the way vscode and renders this is not that important as in the README, and the the extension providers / market places do not need to render it there - those have the "features" tab already which shows the settings and descriptions from package.json in any case).

angelozerr:
@fbricon what do you think about that?

fbricon:
I prefer keeping the settings in the readme for now, it's easier to find.

fbricon:
since the readme is embedded with each release, it's easier to get more accurate information about the settings available in one particular version

GitMensch:
all setup as requested, done with #365

PR #360, opened 2020-11-05
Update Validation.md

GitMensch:
added rewrite sample - not ideal yet, but better than not mentioning it at all

### DIFF
--- a/docs/Validation.md
+++ b/docs/Validation.md
@@ -119,7 +119,7 @@ To bind the XML file `foo.xml` to the XSD file `foo.xsd`, you can use [xml-model
 
 ### XML catalog with XSD
 
-Using an XML catalog can be helpful when you wish to use elements from a namespace but  don't want to set the real path of the XSD file for each XML document.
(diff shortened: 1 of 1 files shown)


datho7561 (changes requested):
Thanks for your contribution! I think this is very helpful. I just have 2 small comments:
 * Could you please edit the commit message to describe what you are changing in the `Validation.md`? I think something like "Improve documentation for using catalogs" could be helpful.
 * See the comment below. If you don't think this change would be help, feel free to ignore the comment
Thanks again!

GitMensch:
I agree completely on both of your points. But I can't edit the message without creating a new commit (I'm only on the end of a browser). Can I suggest you use the changes as template with a message `fixing #360` (so this gets closed on the way)?

GitMensch:
For the change message:

> XML catalog with XSD -> added second example presenting the `rewriteURI` option

angelozerr:
Close for https://github.com/redhat-developer/vscode-xml/pull/361

angelozerr:
@GitMensch thanks for your contribution!

PR #378, opened 2020-12-06
update find-java-home version

### LINKED ISSUE #374: extension does not correctly detect java home correctly without environment variable set
This is more of an annoyance rather than a showstopper. However, having to update JAVA_HOME is annoying because my JAVA is updated automatically. Here's what I know:

- Upon installing the extension, I received a message that JAVA_HOME was not detected.
- Investigating further, vscode-xml appears to use node-find-java version 0.2 to find JAVA home, when JAVA home is not set. This behavior is correct from my POV:

https://github.com/redhat-developer/vscode-xml/blob/f5e2cdc6d2677f703be7ef6ff09bc192644c73c9/src/requirements.ts#L69-L76

- Investigating further, it appears to look in JavaSoft for JAVA_HOME (this is version 0.2, which is in the requirements):

https://github.com/jsdevel/node-find-java-home/blob/1f3dc4e5186dd434859ff6fffb0679feb1e47848/index.js#L61-L65

```
    if(options.allowJre){
      possibleKeyPaths = possibleKeyPaths.concat([
      "SOFTWARE\\JavaSoft\\Java Runtime Environment",
      ]);
    }
```

Here is the structure of my registry:

image

Yet, I still received this error message:

image

Adding this directly to user settings appears to resolve the problem:

image

The question is why this error happened, given that the extension should have found JAVA_HOME by itself. I am not currently set-up to debug vscode extensions, but I may tackle this if I have time.

xoviat:
Fixes #374.

registry logic was incorrect in old version, but fixed in new version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
   "dependencies": {
     "fs-extra": "^8.1.0",
     "expand-home-dir": "^0.0.3",
-    "find-java-home": "0.2.0",
+    "find-java-home": "1.1.0",
     "glob": "^7.1.4",
     "path-exists": "^4.0.0",
     "vscode-languageclient": "^5.2.1"


datho7561:
Still works on Linux, just need to set up my Windows box to test it there...

datho7561 (approved):

datho7561:
Thanks for your contribution!

PR #398, opened 2021-01-28
Improve README to better explain binary

### LINKED ISSUE #397: Improve README documentation surrounding native binary
Remove "running with Java" from the extension description, and explain that a binary version or the Java version can be used.

Explain that the extension downloads the appropriate binary version of LemMinX if the user doesn't have Java.

datho7561:
Closes #397

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## Description
 
-This VS Code extension provides support for creating and editing XML documents, based on the [LemMinX XML Language Server](https://github.com/eclipse/lemminx), running with Java.
+This VS Code extension provides support for creating and editing XML documents, based on the [LemMinX XML Language Server](https://github.com/eclipse/lemminx).
 
 ![Basic features](https://user-images.githubusercontent.com/148698/45977901-df208a80-c018-11e8-85ec-71c70ba8a5ca.gif)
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -44,6 +44,8 @@ or you can read this documentation inside vscode with the command `Open XML Docu
 For running the binary version:
   * Windows, macOS, or Linux, on a x86_64 CPU
     * We don't make a binary specific for Apple ARM (Apple Silicon), but the x86_64 binary seems to work through the Rosetta 2 translation layer.
+  * Java is not required for this version
+  * The binary is automatically downloaded by vscode-xml if it is needed, with no additional action required on the part or the user.
```
datho7561, line 48: oops

rgrunber (commented):
Overall, I'm fine with these changes. I just have a few questions on some other things I noticed.

- Is https://github.com/redhat-developer/vscode-xml/blob/master/docs/Preferences.md#java-home going to be updated separately or is it worth including it in here ?
- The first supported setting mentioned says "Set the Java path required to run the XML Language Server." It needs to be updated but also, I think, given that this is the first setting someone may see when viewing the README, it should probably be reworded to emphasize that this isn't required any longer.

datho7561:
Good catch, one second. I'll amend

rgrunber (approved):
Looks good to me. Feel free to merge this.

PR #404, opened 2021-02-02
Hotfix for the NativeImage pipeline

datho7561:
The publishToMarketPlace property isn't being read properly.


rgrunber (approved):
This falls in line with other usages of 'publishToMarketPlace'. In the main Jenkinsfile it's called directly as a variable and '${params.name}' is reserved for resolving 'name' in a shell script context.

PR #421, opened 2021-02-12
Refactor to improve readability

datho7561:
Use `await` in `extension.ts`. Move some functionality out of `extension.ts`. Move some TypeScript files into subfolders. Fix `.editorconfig`.

Fixes #418

Signed-off-by: David Thompson <davthomp@redhat.com>

PR #427, opened 2021-02-22
Telemetry

### LINKED ISSUE #415: Collect Telemetry data
We want to collect some informations about vscode-xml usage. Let's start with some easy ones :

- lemminx version
- server mode (JVM vs Binary)
- preferBinary (true/false)
- VM name (Oracle/OpenJDK/GraalVM...) + version

Follow the instructions from https://github.com/redhat-developer/vscode-commons/blob/master/INSTRUCTIONS.md

datho7561:
Implement telemetry. Please refer to `USAGE_DATA.md` to see which data is collected.
Closes #415

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,9 @@
     "Formatters",
     "Snippets"
   ],
+  "extensionDependencies": [
+    "redhat.vscode-commons"
+  ],
   "devDependencies": {
     "@types/fs-extra": "^8.0.0",
     "@types/node": "^10.14.16",
@@ -74,6 +77,7 @@
     "webpack-cli": "^3.3.12"
   },
   "dependencies": {
+    "@redhat-developer/vscode-redhat-telemetry": "0.0.13",
     "expand-home-dir": "^0.0.3",
     "find-java-home": "1.1.0",
     "fs-extra": "^8.1.0",
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -1,0 +1,46 @@
+import { getTelemetryService, TelemetryEvent, TelemetryService } from "@redhat-developer/vscode-redhat-telemetry/lib";
+
+/**
+ * Wrap vscode-redhat-telemetry to suit vscode-xml
+ */
+export namespace Telemetry {
+
+  export const OPEN_JAVA_DOWNLOAD_LINK_EVT: string = "open_java_download_link";
+  export const SETTINGS_EVT: string = "settings";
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -34,7 +34,6 @@
   ],
   "main": "./dist/extension",
   "scripts": {
-    "preinstall": "npx npm-force-resolutions",
```
fbricon, line 37: why removing it?

datho7561: The `package-lock.json` format changed in npm 7, so it npm-force-resolutions doesn't work correctly: https://stackoverflow.com/a/65014694.

I shouldn't have made this change, since this likely will cause issues on Jenkins. I'll downgrade my npm so that it will work on my machine.

datho7561:
Going to rebase onto #421 once its merged, marking this PR as ready for review.

src/telemetry.ts
```diff
@@ -0,0 +1,45 @@
...
+   * @param data the telemetry data
+   * @throws Error if the telemetry service has not been started yet
+   */
+  export async function sendTelemetry(eventName: string, data?: any): Promise<void> {
+    if (!_telemetryManager) {
+      throw new Error("The telemetry service for vscode-xml has not been started yet");
+    }
+    await _telemetryManager.send({
+      name: eventName,
+      type: 'trace',
```
fbricon, line 40: track

rgrunber:
I've observed that if I replace the [default segment](https://github.com/redhat-developer/vscode-commons/blob/master/package.json#L9) key with the debug one (ie. always use debug), rebuild vscode-commons, and install it along with this vscode-xml change then events are always logged, and it's working as expected.

fbricon:
> I've observed that if I replace the [default segment](https://github.com/redhat-developer/vscode-commons/blob/master/package.json#L9) key with the debug one (ie. always use debug), rebuild vscode-commons, and install it along with this vscode-xml change then events are always logged, and it's working as expected.
> 
> If not, I only seem to see events recorded when I'm debugging.

Yes, as explained in https://github.com/redhat-developer/vscode-commons/blob/master/INSTRUCTIONS.md#optional-add-a-custom-segment-key-in-packagejson-file

fbricon:
Can we track binary downloads (and failures)?

rgrunber:
I think we can merge this when ready. I can confirm that success/failure do get logged.

datho7561:
Okay. I am ready to merge if you approve the PR.

rgrunber (approved):
Looks good. Let's get this in.

PR #430, opened 2021-02-22
Display a notification when the language server crashes

### LINKED ISSUE #409: No error is reported to the user when the binary fails to start
I installed [0.15.0-108](https://download.jboss.org/jbosstools/vscode-xml/staging/vscode-xml-0.15.0-108.vsix) on my Mac, set preferBinary=true, and after downloading the binary, the server fails to start but no error message pops up to the user.

You need to open the XML output view to see:
```
[Error - 9:17:15 AM] Connection to server is erroring. Shutting down server.
[Error - 9:17:15 AM] Connection to server is erroring. Shutting down server.
[Error - 9:17:15 AM] Connection to server is erroring. Shutting down server.
```

datho7561:
Closes #409

### DIFF
--- a/src/client/clientErrorHandler.ts
+++ b/src/client/clientErrorHandler.ts
@@ -1,0 +1,39 @@
+import { window } from "vscode";
+import { CloseAction, ErrorAction, ErrorHandler, Message } from "vscode-languageclient";
+
+/**
+ * An error handler that restarts the language server,
+ * unless it has been restarted 5 times in the last 3 minutes
+ *
+ * Adapted from [vscode-java](https://github.com/redhat-developer/vscode-java)
+ */
+export class ClientErrorHandler implements ErrorHandler {
+
+  private restarts: number[];
+  private name: string;
+
+  constructor(name: string) {
+    this.name = name;
+    this.restarts = [];
+  }
+
+  error(_error: Error, _message: Message, _count: number): ErrorAction {
+    return ErrorAction.Continue;
+  }
+
+  closed(): CloseAction {
+    this.restarts.push(Date.now());
+    if (this.restarts.length < 5) {
+      return CloseAction.Restart;
+    } else {
+      const diff = this.restarts[this.restarts.length - 1] - this.restarts[0];
+      if (diff <= 3 * 60 * 1000) {
+        window.showErrorMessage(`The ${this.name} language server crashed 5 times in the last 3 minutes. The server will not be restarted.`);
+        return CloseAction.DoNotRestart;
+      }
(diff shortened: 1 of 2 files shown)


datho7561:
To test (under Linux), I made a file `lemminx-linux` under the `server` directory that contained: `exit`, then made it executable. I then set `xml.server.preferBinary = true`, and ran the extension. A notification about the server crashing should show up.

src/extension.ts
```diff
@@ -13,7 +13,7 @@
 import * as os from 'os';
 import * as path from 'path';
 import { Command, commands, ExtensionContext, extensions, IndentAction, LanguageConfiguration, languages, Position, TextDocument, TextEditor, Uri, window, workspace } from "vscode";
-import { CancellationToken, ConfigurationParams, ConfigurationRequest, DidChangeConfigurationNotification, Executable, ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, MessageType, NotificationType, ReferencesRequest, RequestType, RevealOutputChannelOn, TextDocumentIdentifier, TextDocumentPositionParams } from 'vscode-languageclient';
+import { CancellationToken, ConfigurationParams, ConfigurationRequest, DidChangeConfigurationNotification, Executable, ExecuteCommandParams, ExecuteCommandRequest, LanguageClient, LanguageClientOptions, MessageType, NotificationType, ReferencesRequest, RequestType, RevealOutputChannelOn, TextDocumentIdentifier, TextDocumentPositionParams, InitializationFailedHandler, FoldingRangeRequest, ErrorAction, CloseAction } from 'vscode-languageclient';
```
fbricon, line 16: do you need the InitializationFailedHandler, FoldingRangeRequest imports?

datho7561: Good catch! I removed the imports.

rgrunber (commented):
This works for me, so I think once conflicts are resolved, we should merge. One thing to keep in mind is that in vscode-java the restart behaviour tries to restart a few times after a crash and only stays down when it looks like the crashes are happening too often.

https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L54

Something to maybe consider in the future.

datho7561:
Okay. I resolved the merge conflicts. I agree that we should support attempting to relaunch the server like in vscode-java

datho7561:
I switched over to the method that vscode-java (stop trying to restart the server if the server has crashed 5 times in the last 3 minutes)

rgrunber (approved):
Feel free to merge this when ready. Tried it out and the crash timer seems to be working well.

PR #426, opened 2021-02-18
Add support for `textDocument/linkedEditingRange`

angelozerr:
Add support for `textDocument/linkedEditingRange`

See https://github.com/eclipse/lemminx/issues/987

Signed-off-by: azerr <azerr@redhat.com>

PR #432, opened 2021-03-04
Update vscode-languageclient to 7.0.0

datho7561:
This allows for `textDocument/linkedEditingRange` to be used.

Replaces #426

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3748,7 +3748,8 @@
     "semver": {
       "version": "5.7.1",
       "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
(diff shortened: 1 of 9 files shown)


angelozerr:
I think you can switch your draft PR in real PR.

angelozerr:
I'm really annoying because this PR works good with master of LemMinx but with the Leinked Editing range PR https://github.com/eclipse/lemminx/pull/991 (even if linked editing range is disable) the server crashes (no completion, no validation, etc)

datho7561:
I also get this. I'll play around with this a bit today to see if I can figure out what's going on.

angelozerr (approved):

angelozerr:
This PR works great with LemMinx and with LSP4J 0.9 (master) and 0.11 but not for 0.12. As it's not a problem coming from this PR, I merge it. If 0.12 have a problem we can use 0.11 for linked editing range support.

Thanks @datho7561 !

PR #437, opened 2021-03-11
await the Language Client stopping in deactivate

### LINKED ISSUE #436: Extension doesn't stop after VS Code is closed
Running the extension through the debug in VS Code, you can see that there is still some process running after the Extension development window closes:

Screenshot from 2021-03-11 11-06-44

We probably just need to `await` for the LSP client to close before stopping the extension: https://github.com/redhat-developer/vscode-xml/blob/master/src/extension.ts#L58

datho7561:
Closes #436


angelozerr:
@datho7561 could you fix the conflict please.

angelozerr (approved):

angelozerr:
Thanks @datho7561 !

PR #441, opened 2021-03-24
Add option to prevent self-closing from removing content

### LINKED ISSUE #440: Auto self-closing feature deletes content
Type `/` where the pipe is:

```xml
<foo |

  <baloney></baloney>

</foo>
```

Everything else is deleted:

SelfCloseDeletesContent

Expected result:

```xml
<foo />

  <baloney></baloney>

</foo>
```

datho7561:
Closes #440

Requires eclipse/lemminx#1009

### DIFF
--- a/docs/Preferences.md
+++ b/docs/Preferences.md
@@ -133,6 +133,75 @@ See the [Formatting page](Formatting.md#formatting).
 
 Use `xml.codeLens.enabled` in order to enable or disable code lens. Please see [the Code Lens page](CodeLens.md#code-lens) for more information.
 
+## Tag Auto Closing
+
+The closing tags for elements should automatically be added when typing `>` to complete an opening tag.
+
+For instance, if you type `>` when your cursor is placed at the `|` pipe in the following code:
+
+```xml
+<root>
+  <child|
+</root>
+```
+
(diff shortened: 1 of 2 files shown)


angelozerr:
Please add this settings in https://github.com/redhat-developer/vscode-xml/blob/master/docs/Preferences.md

rgrunber (commented):
I don't think I'm seeing this behaviour, where the orphaned text and orphaned-element end tag are removed.

It does work in the cases you used in the tests and bug.

datho7561:
It only happens when closing the root element; maybe that's why?

rgrunber:
auto-close-on-orphaned-text

The docs seemed to imply that the orphaned content should also be removed, but I'm not seeing that. Let me know if I'm missing something in the explanation.

datho7561:
> The docs seemed to imply that the orphaned content should also be removed, ...

The orphaned content is definitely not removed; I'll fix the documentation.

datho7561:
Okay. I removed the nonsense I wrote and replaced it with actual documentation of the feature.

rgrunber (approved):
Thanks for fixing these up.

PR #443, opened 2021-03-29
Update licence to EPL v2.0

fbricon:


rgrunber (commented):
LGTM. Do we need to update the copyright headers in `extension.ts` and `commands/commandConstants.ts` ? I notice most files don't have it either way.

fbricon:
I updated the headers but don't think we need to touch the others. Well we'll prolly need to make things consistent eventually, but I won't bother today.

rgrunber (approved):

PR #447, opened 2021-03-30
Configure XSL auto close tags to work out of the box

### LINKED ISSUE #446: support for *.xslt files
I am new to VS-Code and installed the extension because I do a lot of XSLT work. The feature I am most looking for is the auto-close. It works out of the box for *.xml but it does nto work for *xslt. In principle, an XSLT file is just another XML file. I did not find any item in the configuration to add file extensions (or languages?) to enable the extension for XSLT as well.

datho7561:
Closes #446

### DIFF
--- a/package.json
+++ b/package.json
@@ -479,6 +479,11 @@
         "editor.autoClosingBrackets": "never",
         "files.trimFinalNewlines": true,
         "files.trimTrailingWhitespace": false
+      },
+      "[xsl]": {
+        "editor.autoClosingBrackets": "never",
+        "files.trimFinalNewlines": true,
+        "files.trimTrailingWhitespace": false
       }
     },
     "commands": [


package.json
```diff
@@ -479,6 +479,9 @@
         "editor.autoClosingBrackets": "never",
         "files.trimFinalNewlines": true,
         "files.trimTrailingWhitespace": false
+      },
+      "[xsl]": {
```
angelozerr, line 483: Even if it's not the scope of the fix, I will do like xml:

```json
"[xsl]": {
        "editor.autoClosingBrackets": "never",
        "files.trimFinalNewlines": true,
        "files.trimTrailingWhitespace": false
      }
```

angelozerr (approved):

PR #449, opened 2021-04-06
Inject VS Code proxy settings into JVM args

### LINKED ISSUE #416: Respect http.proxy and http.proxyAuthorization settings
Proxy support doesn't work out of the box

datho7561:
Interprets the `http.proxyHost` and `http.proxyAuthorization` settings in order to set up the JVM arguments or environment variables in order to configure the proxy for the language server.

JVM arguments are used when running the Java server, and environment variables are used when running the binary server.

This means if the proxy is configured through these settings in VS Code, then vscode-xml will download schemas through the proxy.

While running the Java server, if the proxy is already configured in `xml.server.vmargs`, then the `http.proxyHost` and `http.proxyAuthorization` are ignored.

Limitations:
 * If `http.proxyAuthorization` is used, it is assumed to be [Basic authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#directives), with a username/password pair, since the JVM expects a username/password pair.

Requires eclipse/lemminx#1012

Closes #416

### DIFF
--- a/src/server/binary/binaryServerStarter.ts
+++ b/src/server/binary/binaryServerStarter.ts
@@ -5,12 +5,13 @@ import * as https from 'https';
 import * as os from 'os';
 import * as path from 'path';
 import { Readable } from 'stream';
-import { ExtensionContext, extensions, window, WorkspaceConfiguration, ProgressOptions, ProgressLocation, Progress, CancellationToken } from "vscode";
+import { ExtensionContext, extensions, ProgressLocation, ProgressOptions, window, WorkspaceConfiguration } from "vscode";
 import { Executable } from "vscode-languageclient";
 import * as yauzl from 'yauzl';
-import { addTrustedHash, getTrustedHashes } from './binaryHashManager';
+import { getProxySettings, getProxySettingsAsEnvironmentVariables, ProxySettings } from '../../settings/proxySettings';
 import { getXMLConfiguration } from "../../settings/settings";
 import { Telemetry } from '../../telemetry';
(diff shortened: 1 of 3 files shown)


src/server/binary/binaryServerStarter.ts
```diff
@@ -26,13 +27,16 @@ export const ABORTED_ERROR: Error = new Error('XML Language Server download canc
...
   let binaryExecutable: Executable;
   return getServerBinaryPath()
     .then((binaryPath: string) => {
       binaryExecutable = {
-        args: [binaryOptions],
-        command: binaryPath
+        args: [binaryArgs],
+        command: binaryPath,
+        options: {
+          env: getProxySettingsAsEnvironmentVariables(getProxySettings())
```
datho7561, line 38: What happens if no proxy is configured?

src/settings/settings.ts
```diff
@@ -1,6 +1,7 @@
 import * as path from 'path';
 import { commands, Extension, extensions, window, workspace, WorkspaceConfiguration } from "vscode";
 import { getVariableSubstitutedAssociations } from "./variableSubstitution";
+import { jvmArgsContainsProxySettings, getProxySettings, ProxySettings, getProxySettingsAsJVMArgs } from './proxySettings';
```
datho7561, line 4: Remove

rgrunber (approved):
I've tested this out and seem like values are being passed to the language server for the Java and Binary cases, so this seems fine to me.

Feel free to merge when ready.

Issue #457, opened 2021-04-20
Build a static binary for Linux


datho7561:
This should allow for running on systems without glibc, such as Alpine

PR #460, opened 2021-04-21
Revert "Update vscode-languageclient to 7.0.0"

datho7561:
This reverts commit c248ffb6bd51ec9b459b07e1828d0e718a2e1b02.

We need to prepare a service release (0.16.1) in order to fix the binary (see #457). The new version of the language client doesn't work with Theia, so for the purpose of the service release, we should roll back the client update.


rgrunber (approved):
LGTM. Seems to work as before. Tested the binary server as well.

PR #468, opened 2021-05-07
Perform a fake checkout of vscode-xml early to prevent rebuild cycles

rgrunber:
- If there are new changes in vscode-xml and the build fails (eg.
  generating a native binary), it never gets to properly checkout
  vscode-xml, and the relevant changes in vscode-xml are marked as new
  in subsequent git polling


datho7561 (approved):
I think this is worth a try to see if it prevents the rebuilds

PR #472, opened 2021-05-10
Use Visual Studio 2019 in NativeImage.jenkins

datho7561:
The executable available at https://go.microsoft.com/fwlink/&LinkId=691126 no longer successfully installs the Microsoft C/C++ compiler. This means that, currently, our Windows build agents don't have the Microsoft C/C++ compiler installed. One way to overcome this obstacle is to install a newer version of the Microsoft C/C++ compiler. The newer version changes the directory structure of where things are installed. This PR changes the paths that reference the Microsoft C/C++ compiler installation so that they point to the corresponding directories for the 2019 version.

### DIFF
--- a/NativeImage.jenkins
+++ b/NativeImage.jenkins
@@ -69,8 +69,8 @@ pipeline {
             bat """
               pushd .
               setlocal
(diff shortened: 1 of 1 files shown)


datho7561:
Requires the builder/agents to have MSVC 2019 installed

rgrunber (approved):
LGTM. This has already been tested (with temporary builds), so this would just be making the change official.

PR #490, opened 2021-06-01
Automatically closing tags in multi-cursor mode only closes one tag

### LINKED ISSUE #225: Automatically closing tags in multi-cursor mode only closes one tag
## Description

In multi-cursor mode, the autocompletion only completes one XML Tag.

## Steps to reproduce

1. Set `"xml.completion.autoCloseTags": true`.
2. Type the following (pipes indicate multiple cursor locations):

```
    <lib|
    <lib|
    <lib|
    <lib|
```

3. Type `>`

### Expected
I expect the document to look like this:

```
    <lib>|</lib>
    <lib>|</lib>
    <lib>|</lib>
    <lib>|</lib>
```

### Actual 

Instead, the document looks like this:

```
    <lib>|</lib>
    <lib>|
    <lib>|
    <lib>|
```

angelozerr:
Automatically closing tags in multi-cursor mode only closes one tag

Fixes #225

### DIFF
--- a/src/client/tagClosing.ts
+++ b/src/client/tagClosing.ts
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
- * 
+ *
  *  Retrieved from: https://github.com/Microsoft/vscode/blob/f707828426bd87e88c17d2da34f2ceed0019d8bd/extensions/html-language-features/client/src/tagClosing.ts
  *--------------------------------------------------------------------------------------------*/
 'use strict';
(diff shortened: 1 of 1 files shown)


angelozerr:
The PR was easy to do, I have just copied/pasted code from https://github.com/microsoft/vscode/blob/main/extensions/html-language-features/client/src/tagClosing.ts

The only thing that I noticed is that the tagComplete coming from language server should return only string instead of string+range.

I will create a PR on LemMinx side to fix that, but you can play with the PR now, it should work.

datho7561:
I decided to test this use case:

multi-cursor-close

We could try to support adding the correct close tag by passing an array of cursors in the `xml/closeTag` request. Another way might be to make multiple requests to LemMinX (i.e. request close for first cursor, apply the result, request close for second cursor, apply...)

Then again, it might not be worthwhile to support this use case, since you expect the same text to appear at all the cursors when using multi cursor mode.

angelozerr:
@datho7561 for the moment I would like to have the same behavior than HTML language feature. Your idea seems great, but perhaps we should discuss about that with vscode html language service team to have the same code tagClosing.ts

fbricon:
@datho7561 nice catch! However I think we can fix it later, as you said, the common use case, from a user standpoint, is to multi-close the same tags.

Unrelated: to review the PR while ignoring whitespace changes, append ?w=1 to the files url: https://github.com/redhat-developer/vscode-xml/pull/490/files?w=1

angelozerr:
> Unrelated: to review the PR while ignoring whitespace changes, append ?w=1 to the files url: https://github.com/redhat-developer/vscode-xml/pull/490/files?w=1

The code is just a copy/paste from html language feature of vscode.

src/client/tagClosing.ts
```diff
@@ -14,87 +14,73 @@ export interface AutoCloseResult {
...
+    let lastCharacter = lastChange.text[lastChange.text.length - 1];
+    if (lastChange.rangeLength > 0 || lastCharacter !== '>' && lastCharacter !== '/') {
+      return;
+    }
+    let rangeStart = lastChange.range.start;
+    let version = document.version;
+    timeout = setTimeout(() => {
+      let position = new Position(rangeStart.line, rangeStart.character + lastChange.text.length);
+      tagProvider(document, position).then(result => {
+        const text = result?.snippet;
```
datho7561, line 64: This means that the functionality described in https://github.com/redhat-developer/vscode-xml/issues/440 no longer works, since that functionality depends on using the returned range to determine what should be deleted.

src/client/tagClosing.ts
```diff
@@ -14,87 +14,89 @@ export interface AutoCloseResult {
...
+        if (text && isEnabled) {
+          let activeEditor = window.activeTextEditor;
+          if (activeEditor) {
+            let activeDocument = activeEditor.document;
+            if (document === activeDocument && activeDocument.version === version) {
+              let selections = activeEditor.selections;
+              if (selections.length && selections.length > 1 && selections.some(s => s.active.isEqual(position))) {
```
datho7561, line 71: `selections` is an array that is guaranteed to have a `length`, so you shouldn't need to check the first condition.

datho7561 (changes requested):
Looks good, and works great! Just one small change

src/client/tagClosing.ts
angelozerr: good catch. fixed

datho7561 (approved):

PR #494, opened 2021-06-04
Bind XSD, DTD with CodeLens

### LINKED ISSUE #395: Bind XSD, DTD with CodeLens
A very useful feature for an XML IDE would be a button or action to associate an XML document that is currently open in VS Code with a schema file.

After clicking a button, a file dialog would open which would allow to select an XSD file from the file system. Then, your XML extension would insert into the XML document the correct attributes or processing instructions needed to make this association.

The reason why that would be useful is that there are several ways to associate an XML document with an XSD file:

https://github.com/redhat-developer/vscode-xml/blob/master/docs/Validation.md#validation-with-xsd-grammar

all of them cumbersome, and hard to remember. Especially because you also have to pay close attention and be aware that it makes a difference whether the XML document to be associated has a namespace or not.

Oxygen has functionality like this and I find it very useful.

angelozerr:
Bind XSD, DTD with CodeLens

Fixes #395

### DIFF
--- a/src/commands/commandConstants.ts
+++ b/src/commands/commandConstants.ts
@@ -11,52 +11,78 @@
 'use strict';
 
 /**
- * Commonly used commands
+ * VScode client commands.
  */
-export namespace CommandConstants {
+export namespace ClientCommandConstants {
 
-    /**
-     * Auto close tags
-     */
-    export const AUTO_CLOSE_TAGS = 'xml.completion.autoCloseTags';
+  /**
+   * Show XML references
+   */
+  export const SHOW_REFERENCES = 'xml.show.references';
 
-    /**
-     * Show XML references
-     */
-    export const SHOW_REFERENCES = 'xml.show.references';
+  /**
+   * Show editor references
+   */
+  export const EDITOR_SHOW_REFERENCES = 'editor.action.showReferences';
 
-    /**
-     * Show editor references
-     */
-    export const EDITOR_SHOW_REFERENCES = 'editor.action.showReferences';
+  /**
+   * Reload VS Code window
+   */
+  export const RELOAD_WINDOW = 'workbench.action.reloadWindow';
 
-    /**
-     * Reload VS Code window
-     */
-    export const RELOAD_WINDOW = 'workbench.action.reloadWindow';
+  /**
+   * Open settings command
+   *
+   * A `settingId: string` parameter can be optionally provided
+   */
+  export const OPEN_SETTINGS = 'xml.open.settings';
 
-    /**
-     * Open settings command
-     *
-     * A `settingId: string` parameter can be optionally provided
-     */
-    export const OPEN_SETTINGS = 'xml.open.settings';
+  /**
+   * Render markdown string to html string
+   */
+  export const MARKDOWN_API_RENDER = 'markdown.api.render';
 
-    /**
-     * Render markdown string to html string
-     */
-    export const MARKDOWN_API_RENDER = 'markdown.api.render';
+  export const OPEN_DOCS = 'xml.open.docs';
 
-    export const OPEN_DOCS = "xml.open.docs";
+  /**
+   * Commands to revalidate files with an LSP command on the XML Language Server
+   */
+  export const VALIDATE_CURRENT_FILE = 'xml.validation.current.file';
 
-    export const OPEN_DOCS_HOME = "xml.open.docs.home";
+  export const VALIDATE_ALL_FILES = 'xml.validation.all.files';
 
-    /**
-     * VSCode client command to executes an LSP command on the XML Language Server
-     */
-    export const EXECUTE_WORKSPACE_COMMAND = "xml.workspace.executeCommand";
+  export const OPEN_DOCS_HOME = 'xml.open.docs.home';
 
-    export const VALIDATE_CURRENT_FILE = "xml.validation.current.file";
+  /**
+   * VSCode client commands to open the binding wizard to bind a XML to a grammar/schema.
+   */
+  export const OPEN_BINDING_WIZARD = 'xml.open.binding.wizard';
 
-    export const VALIDATE_ALL_FILES = "xml.validation.all.files";
+  /**
+   * Client command to execute an XML command on XML Language Server side.
+   */
+  export const EXECUTE_WORKSPACE_COMMAND = 'xml.workspace.executeCommand';
+}
+
+/**
+ * XML Language Server commands.
+ */
+export namespace ServerCommandConstants {
+
+  /**
+   * Auto close tags
+   */
+  export const AUTO_CLOSE_TAGS = 'xml.completion.autoCloseTags';
+
+  /**
+   * Commands to revalidate files with an LSP command on the XML Language Server
+   */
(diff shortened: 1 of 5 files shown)


angelozerr:
This PR requires the LemMinx PR https://github.com/eclipse/lemminx/pull/1049

angelozerr:
Here a little demo with XSD, DTD codelens binding

CodeLensBindingDemo

AlexXuChen:
Will this only bound the file from the absolute path on the local machine? (e.g. `c:\Users\..`) If the developer is making a xml front end in production, and forgets to change the path, they will likely run into problems. Is this feature intended just for local development?

datho7561:
I think Angelo mentioned that he was going to try to implement relative paths eventually

angelozerr:
I have improved the PR:

 * relative path should be now supported.
 * xsi:schemaLocation is generated now wgen XSD defined a targetNamespace.

datho7561:
Here is a more detailed explanation of the performance issue I am facing (under Linux):

When the open file dialog opens, my cursor is set to the index finger pointing up style cursor. When I attempt any mouse interaction with the open file dialog or the instance of VS Code that is debugging the extension, nothing happens. While this is happening, keyboard interaction with the dialog window works fine. I can get mouse interaction with the open file dialog to work again if I click on a text entry field in another window, then go back tot he open file dialog.

I will investigate if this happens with all extensions by coding a quick demo extension, and file upstream in VS Code/Electron if this is the case.

datho7561:
I haven't gotten any further in to figuring out why the Open File dialog doesn't work as expected under Linux.

angelozerr:
> I haven't gotten any further in to figuring out why the Open File dialog doesn't work as expected under Linux.

Could you retry please.

datho7561:
I still have the same issue.

datho7561:
If I use the option `"files.simpleDialog.enable": true`, this PR works fine, and it gives a file picker similar to the Ctrl+P one. Maybe we can temporarily enable/disable this when the command is run? (It will be a bit hacky though).

datho7561:
This appears to be the same issue, but I can no longer reproduce the issue with the steps given in the issue: https://github.com/microsoft/vscode/issues/62704

datho7561:
See microsoft/vscode#126088

src/commands/commandConstants.ts
```diff
@@ -52,11 +52,18 @@ export namespace CommandConstants {
     export const OPEN_DOCS_HOME = "xml.open.docs.home";
 
     /**
-     * VSCode client command to executes an LSP command on the XML Language Server
+     * VSCode client commands to revalidate files with an LSP command on the XML Language Server
```
datho7561, line 55: Move this comment down to the line before `xml.validation.current.file`.

src/commands/registerCommands.ts
```diff
@@ -108,4 +109,75 @@ async function registerValidationCommands(context: ExtensionContext): Promise<vo
...
+    const documentURI = Uri.parse(uriString);
+    // Open a dialog to select the XSD, DTD to bind.
+    const options = createDialogOptions(bindingType);
+    const fileUri = await window.showOpenDialog(options);
+    if (fileUri && fileUri[0]) {
+      // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
+      const identifier = TextDocumentIdentifier.create(documentURI.toString());
+      const grammarURI = fileUri[0];
+      try {
+        const result = await commands.executeCommand(CommandConstants.EXECUTE_WORKSPACE_COMMAND, CommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
```
datho7561, line 131: You don't need to call `EXECUTE_WORKSPACE_COMMAND`, you should be able to call `ASSOCIATE_GRAMMAR_INSERT` directly

datho7561 (changes requested):
Looks pretty good. The issue I found with the PR is an upstream (VS Code or Electron) issue.

src/commands/commandConstants.ts
angelozerr: fixed

src/commands/registerCommands.ts
angelozerr: Are you sure? `EXECUTE_WORKSPACE_COMMAND `means that it's a XML command (from the server). I did a refactoring by splittung CommandConstants to CLientCommandConstants and ServerCommandConstants to avoid this confusion.

datho7561: I deleted the first parameter (`ClientCommandConstants.EXECUTE_WORKSPACE_COMMAND`) in the `commands.executeCommand` function call, and the bind to file functionality still works.

src/commands/registerCommands.ts
```diff
@@ -87,25 +88,83 @@ async function registerCodeLensCommands(context: ExtensionContext, languageClien
...
+      }
+    };
+
+    const fileUri = await window.showOpenDialog(options);
+    if (fileUri && fileUri[0]) {
+      // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
+      const identifier = TextDocumentIdentifier.create(documentURI.toString());
+      const grammarURI = fileUri[0];
+      try {
+        const result = await commands.executeCommand(ClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
```
datho7561, line 156: I still don't think you need `EXECUTE_WORKSPACE_COMMAND`, it works fine if you directly call the `ASSOCIATE_GRAMMAR_INSERT` command.

datho7561 (changes requested):

src/commands/registerCommands.ts
angelozerr: Let me experiment that i dont understand how it could work?

angelozerr:
@datho7561 since de open a combo with binding type encore opening file dialogues is it fixing tour issue with dialog freeze on Linux?

src/commands/registerCommands.ts
datho7561: Any chance that you could address this?

angelozerr: Indeed it works, I updated it. It works because LemMinX returns the ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT (xml.associate.grammar.insert) as workspace command:

```
[Trace - 17:10:40] Received request 'client/registerCapability - (19)'.
Params: {
    "registrations": [
        {
            "id": "157b25db-dfea-4a42-bdea-953be34d71b3",
            "method": "workspace/executeCommand",
            "registerOptions": {
                "commands": [
                    "xml.associate.grammar.insert",
                    "xml.validation.current.file",
                    "xml.validation.all.files"
                ]
            }
        }
    ]
}

```

I tried to the same thing for `xml.validation.current.file` but it doesn't work. I suppose it's because we use the same command ID in the package.json to register the command than xml.validation.current.file.

Could you create an issue about that please.

datho7561 (approved):

PR #497, opened 2021-06-07
Optimize images size

fbricon:
vsix size went from 15.3MB to 13.4MB


fbricon:
tuning .vscodeignore shaves off another 200KB

.vscodeignore
```diff
@@ -2,4 +2,10 @@
...
 tsconfig.json
-webpack.config.js
\ No newline at end of file
+webpack.config.js
+gulpfile.js
+.editorconfig
+CONTRIBUTING.md
+Jenkinsfile
+*.jenkins
+images/
```
datho7561, line 11: We will likely need to include `images/` in the `.vsix`, since these images are linked in the `README.md`. The `README.md` is used in the extension marketplace. There are differences between https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml and the `README.md` from the repo. This means the marketplace is likely using the `README.md` and images that are bundled in the `.vsix` instead of from GitHub. This means if we don't bundle the images, then they won't appear on the marketplace.

datho7561 (changes requested):

.vscodeignore
fbricon: the images/ folder at the root only contains images used by contributing.md. 
Readme uses URLs to the GH repo. 

I tried the embedded doc after installing the vsix and it worked (but you should check it anyway)

datho7561: Okay. Sorry about that.

datho7561 (approved):

PR #501, opened 2021-06-08
Link to docs when binary server download fails due to proxy

### LINKED ISSUE #492: Better handle failed  binary downloads due to "407 Proxy Authentication Required"
Telemetry shows a number of users are failing to download the binary because of "Server binary download failed: status code 407".  407 error means Proxy Authentication Required. When such errors are encountered, we could display a popup with (links to) instructions to configure proxy settings for vscode-xml.

datho7561:
Closes #492

### DIFF
--- a/docs/Proxy.md
+++ b/docs/Proxy.md
@@ -1,0 +1,39 @@
+# Proxy
+
+This page explains how to get vscode-xml to work through a proxy.
+
+## Prerequisites
+
+If you want to use the binary language server,
+then you will need to make sure that your proxy allows connections to jboss.org
+since the binary language server is downloaded from that website.
+You will also need to make sure that you can access any online XML schemas that you want to work with through the proxy.
+
+## Setting up the proxy in VS Code
+
+VS Code provides a setting called `http.proxySupport`.
+If you set it to `override`, VS Code will attempt to rewrite the requests from extensions to go through the proxy that is configured for VS Code.
+However, this method sometimes does not work correctly.
+
+The tested way to get vscode-xml to use your proxy is to set the following settings:
+ * `http.proxy`:
+   The address at which the proxy can be accessed.
+   As an example, use `"http://localhost:3128"` for a proxy running on your machine on port 3128.
+ * `http.proxyAuthorization`:
+   The authorization header to use for the proxy requests.
+   This is only needed when the proxy requires authorization.
+   This follows the form `"[type] [credientials]"`,
+   where `type` is the authorization type that the proxy uses,
+   and `credentials` are your credentials for accessing the proxy.
+   The format of the `credentials` is different for each authorization type.
+   For instance, `"Basic dXNlcm5hbWU6cGFzc3dvcmQ="` is the authorization header
+   for a proxy that uses Basic authorization,
+   where your username is `username` and your password is `password`.
+   The credentials, `"dXNlcm5hbWU6cGFzc3dvcmQ="`, is `username:password` encoded in base64,
+   which is the format that is expected for Basic authorization.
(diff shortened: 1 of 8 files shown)


src/server/binary/binaryServerStarter.ts
```diff
@@ -413,3 +425,7 @@ async function acceptBinaryDownloadResponse(response: http.IncomingMessage): Pro
     response.pipe(serverBinaryFileStream);
   });
 }
+
+async function openProxyDocumentation(): Promise<void> {
```
fbricon, line 429: can we track the open proxy doc event?

datho7561: Okay, I've added a telemetry event for clicking the link. What do you think of the name of the event?

fbricon:
@snjeza I know you frequently test proxy issues, can you please check this one works as expected?

src/server/binary/binaryServerStarter.ts
datho7561: I forgot to document the event in the USAGE_DATA.md

rgrunber (commented):
I set up a basic local proxy using tinyproxy that required authentication and configured it in vscode, leaving out the authentication credentials, since that's what a 407 seems to be about. I saw the popup to configure it correctly as expected.

However, is the telemetry request supposed to be cached for sending at a later time, once the user fixes their proxy settings ? Requests to api.segment.io would get dropped as well without authentication.

datho7561:
It looks like its queued https://github.com/redhat-developer/vscode-redhat-telemetry/blob/main/src/utils/telemetryEventQueue.ts. I'll research if VS Code needs to restart to apply proxy settings.

datho7561:
I added a new page in the documentation that explains how to set up the proxy to work with the extension. It has more specific instructions than the documentation that I was linking to before. Hopefully the fact that its available offline is also helpful.

Its odd that the "download failed" events reach Segment when the proxy is not configured correctly. I guess that would mean whatever proxy is being used allows communication with Segment, but not jboss.org.

docs/Proxy.md
```diff
@@ -0,0 +1,25 @@
...
+
+VS Code provides a setting called `http.proxySupport`.
+If you set it to `override`, VS Code will attempt to rewrite the requests from extensions to go through the proxy that is configured for VS Code.
+However, this method sometimes does not work correctly.
+
+The tested way to get vscode-xml to use your proxy is to set the following settings:
+ * `http.proxy`
+ * `http.proxyAuthorization`, if you use an authenticated proxy
+
+Once these settings are set,
```
rgrunber, line 22: Does LemMinX respect the VS Code proxy settings ? I see the JVM has a separate set of properties.

datho7561: vscode-xml will check if the user configured JVM args has the parameters to set up the proxy. If they don't, vscode-xml converts the values of these settings into the necessary JVM args and appends them to the existing args.

docs/Proxy.md
```diff
@@ -0,0 +1,25 @@
...
+since the binary language server is downloaded from that website.
+You will also need to make sure that you can access any online XML schemas that you want to work with through the proxy.
+
+## Setting up the proxy in VS Code
+
+VS Code provides a setting called `http.proxySupport`.
+If you set it to `override`, VS Code will attempt to rewrite the requests from extensions to go through the proxy that is configured for VS Code.
+However, this method sometimes does not work correctly.
+
+The tested way to get vscode-xml to use your proxy is to set the following settings:
```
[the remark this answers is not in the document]
datho7561: I think it is. I'll add a description for both settings.

docs/Proxy.md
rgrunber: Right. I can see that in settings/proxySettings.ts.

rgrunber (approved):
Feel free to merge this. We can address additional issues (if any) in later commits.

PR #502, opened 2021-06-08
Fix links to parent folder in internal docs

### LINKED ISSUE #499: Internal documentation doesn't support links to pages in parent directories
If you have an internal documentation page in a folder nested under `docs` (i.e. `docs/Features/XMLFeatures.md`) and this Markdown page has a link to a Markdown page in the parent folder (i.e. `../CodeLens.md`, meaning `docs/CodeLens.md` ), then this link will not work properly in the internal documentation.

datho7561:
If you link to a file in the parent directory from a markdown file in a nested folder, the link will now function correctly in the embedded documentation. The embedded documentation is the documentation that's available through the "XML: Open XML Documentation" command.

Fixes #499

### DIFF
--- a/src/markdownPreviewProvider.ts
+++ b/src/markdownPreviewProvider.ts
@@ -55,7 +55,10 @@ class MarkdownPreviewProvider implements Disposable {
             // where $1, $2, $3 are non empty strings that are then passed to the replace function
             markdownString = markdownString.replace(/\[([^\]]+)\]\(([^#\)]+)#([^)]*)\)/g,
                     (_match: string, linkText: string, page: string, section: string) => {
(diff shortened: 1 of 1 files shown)


AlexXuChen (approved):
Everything looks good to me, just wondering why `_match` is there but not used.

datho7561:
> ...just wondering why `_match` is there but not used.

I use a regex to match Markdown-style links. I use [capture groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges#using_groups) to capture the text for the `linkText`, `page`, and `section`. The callback function is passed the entire matched string (which I don't care about), then the contents of each capture group (which I use to build the new link that works in the internal docs).

PR #511, opened 2021-06-16
Prevent `docs/images` from being excluded from .vsix

### LINKED ISSUE #510: Adapt `.vscodeignore` to the new path matching strategy
vsce interprets the contents of `.vscodeignore` differently as of a recent change. This means that everything under `docs/images` is excluded from the .vsix (likely because of the rule `images/`). Adapt `.vscodeignore` so that everything that needs to be included is included.

datho7561:
Fixes #510

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -9,3 +9,4 @@ CONTRIBUTING.md
 Jenkinsfile
 *.jenkins
 images/
+!docs/images


rgrunber (approved):
Tested with `vsce ls`. Seems to fix the issue.

PR #512, opened 2021-06-16
Add support for workspace trust

### LINKED ISSUE #506: Support workspace trust
See https://github.com/redhat-developer/vscode-java/issues/1926 for reference.

Current options/capabilities under consideration for restriction in an untrusted workspace :

- `xml.java.home` & `xml.server.vmargs`
  - We partly restrict these. It can be set in the workspace, but we detect it and prompt the user for `xml.java.home` and restrict javaagent for `xml.server.vmargs`
-  ~`xml.server.binary.{path,args,trustedHashes}`~
   - We already restrict these and can't be set by the workspace
- `xml.validation.resolveExternalEntities`
  - We don't restrict this at all, so we could do that, and certainly restrict it with workspace trust
- Downloading the XML server binary : should we restrict this entirely when untrusted ?
- `xml.validation.enabled`
  - Should this be set to false when a workspace is untrusted ?

rgrunber:
- Restrict workspace access to some settings in untrusted mode
- Fixes #506

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,17 @@
     "osx-x86_64": "https://download.jboss.org/jbosstools/vscode/snapshots/lemminx-binary/LATEST/lemminx-osx-x86_64.zip",
     "win32": "https://download.jboss.org/jbosstools/vscode/snapshots/lemminx-binary/LATEST/lemminx-win32.zip"
   },
+  "capabilities": {
+    "untrustedWorkspaces": {
+      "supported": "limited",
+      "restrictedConfigurations": [
+        "xml.extension.jars",
+        "xml.java.home",
+        "xml.server.vmargs",
+        "xml.validation.resolveExternalEntities"
+      ]
+    }
+  },
   "engines": {
     "vscode": "^1.37.0"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,6 +23,16 @@
     "osx-x86_64": "https://download.jboss.org/jbosstools/vscode/snapshots/lemminx-binary/LATEST/lemminx-osx-x86_64.zip",
     "win32": "https://download.jboss.org/jbosstools/vscode/snapshots/lemminx-binary/LATEST/lemminx-win32.zip"
   },
+  "capabilities": {
+    "untrustedWorkspaces": {
+      "supported": "limited",
+      "restrictedConfigurations": [
+        "xml.java.home",
+        "xml.server.vmargs",
+        "xml.validation.resolveExternalEntities"
```
datho7561, line 32: We should also restrict the `xml.extension.jars` configuration.

datho7561 (changes requested):

datho7561 (approved):

PR #520, opened 2021-06-23
Fix extension not starting in vscode < 1.55

fbricon:
Update vscode-redhat-telenetry to 0.2.0, to fix

```
Activating extension 'redhat.vscode-xml' failed: Cannot read property 'id' of undefined.
```

This affects vscode API < 1.55 (Che/Theia are at 1.50)


datho7561:
vscode-xml won't run in Theia anyways, since this release requires the language client version 7, which requires being compatible with the 1.52 API. However, it would be nice to support VS Code 1.52+ and Theia once it's compatible with the new language client version.

datho7561 (approved):

PR #522, opened 2021-06-23
Document auto tag renaming in README

datho7561:
Add a quick explanation on how to enable it. Fix link that explains the old method of tag renaming.

### DIFF
--- a/docs/Features/XMLFeatures.md
+++ b/docs/Features/XMLFeatures.md
@@ -14,7 +14,18 @@ Using `/` in an opening tag will auto close the tag.
 
 Linked editing is supported, allowing for simultaneous changes an opening and closing tag pair.
 
-To enable this feature, the setting `editor.linkedEditing` must be set to `true` in the `settings.json.
+This feature is enabled by default for XML and XSL files.
+
+If you would like to disable it, add the following to your `settings.json`:
+
+```json
+"[xml]" : {
+  "editor.linkedEditing": false
+},
+"[xsl]" : {
+  "editor.linkedEditing": false
+}
+```
 
 ![Linked Editing](../images/Features/LinkedEditingXML.gif)
 
--- a/README.md
+++ b/README.md
@@ -14,13 +14,14 @@ This VS Code extension provides support for creating and editing XML documents, 
 
   * Syntax error reporting
   * General code completion
-  * [Auto-close tags](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Features.md#xml-tag-auto-close)
+  * [Auto-close tags](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Features/XMLFeatures.md#xml-tag-auto-close)
   * Automatic node indentation
   * Symbol highlighting
   * Document folding
   * Document links
   * [Document symbols and outline](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Symbols.md)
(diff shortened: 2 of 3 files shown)


AlexXuChen (changes requested):
The "when you enable `editor.linkedEditing`" could be removed to keep it consistent with the other links that don't list their prerequisites. It is also mentioned already in the `XMLFeatures.md`

fbricon:
> The "when you enable `editor.linkedEditing`" could be removed to keep it consistent with the other links that don't list their prerequisites. It is also mentioned already in the `XMLFeatures.md`

all other features are enabled by default. I think it's worth highlighting that particular requirement. People will likely see the readme, the doc, not so much

datho7561:
Okay, I've changed it back. Also fixed a small issue in the Features page documentation.

fbricon (approved):

AlexXuChen:
> all other features are enabled by default. I think it's worth highlighting that particular requirement. People will likely see the readme, the doc, not so much

Then to keep it consistent, could we potentially add what prerequisite(s) each feature needs enabled to work, similar to how David listed it?

fbricon:
> > all other features are enabled by default. I think it's worth highlighting that particular requirement. People will likely see the readme, the doc, not so much
> 
> Then to keep it consistent, could we potentially add what prerequisite(s) each feature needs(if any) enabled to work, similar to how David listed it?

sure, if it's disabled by default.

datho7561:
Wait a second.... I think I've got a solution

datho7561:
I've enabled it by default.

datho7561:
@fbricon @AlexXuChen do you think this strategy is good, or should I just document how to enable the feature? This is the only feature in the list that requires action from the use to enable.

AlexXuChen:
> @fbricon @AlexXuChen do you think this strategy is good, or should I just document how to enable the feature? This is the only feature in the list that requires action from the use to enable.

I don't see a particular case where enabling this by default will cause a user to have a bad experience (i.e. if they are changing the starting tag, it is very likely that they will change the ending tag as well, any other use case would mean they delete the tag entirely, which isn't an issue in this case).

I do believe however that we should consider a strategy for when we actually do implement an option that is based on preference (disabled by default) and how we can consistently document this in the `README`.

docs/Features/XMLFeatures.md
```diff
@@ -14,7 +14,18 @@ Using `/` in an opening tag will auto close the tag.
...
 
-To enable this feature, the setting `editor.linkedEditing` must be set to `true` in the `settings.json.
+This feature is enabled by default for XML and XSL files.
+
+If you would like to disable it, add the following to your `settings.json`:
+
+```json
+"[xml]" : {
+  "editor.linkedEditing: false
+}
```
datho7561, line 24: eek I forgot a comma

docs/Features/XMLFeatures.md
```diff
@@ -14,7 +14,18 @@ Using `/` in an opening tag will auto close the tag.
...
 Linked editing is supported, allowing for simultaneous changes an opening and closing tag pair.
 
-To enable this feature, the setting `editor.linkedEditing` must be set to `true` in the `settings.json.
+This feature is enabled by default for XML and XSL files.
+
+If you would like to disable it, add the following to your `settings.json`:
+
+```json
+"[xml]" : {
+  "editor.linkedEditing: false
```
datho7561, line 23: And the closing quotes too! goodness

AlexXuChen (approved):

PR #525, opened 2021-06-24
Linked editing no longer enabled by default

datho7561:
Document that its not enabled by default. Explain how to enable it in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ This VS Code extension provides support for creating and editing XML documents, 
 
 ## Features
 
+| Regular font       | *Italics* font                              |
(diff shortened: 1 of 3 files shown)


rgrunber (approved):
I'm fine with this. If we end up having more features we list here that are disabled, then I think we should remove any explanation and just ensure we have a link to the docs that directly explain what options are needed to enable.

Also don't forget to update the date listed for 0.17.0 release on the CHANGELOG. We'll probably have a better idea after investigating the shutdown/exit issue.

datho7561:
I'll look into updating the release date on the PR that addresses the shutdown issue.

PR #526, opened 2021-06-24
Manually send the shutdown and exit requests

datho7561:
Hack to avoid microsoft/vscode-languageserver-node#755


datho7561:
This approach does not work

PR #530, opened 2021-06-25
Support `shouldLanguageServerExitOnShutdown` capability

datho7561:
Supports the capability, which means that language server shuts down when `shutdown` is received instead of when `exit` is received.

See eclipse/lemminx#1070


datho7561:
Need to change package.json to lemminx 0.17.1

rgrunber (approved):
LGTM.

PR #544, opened 2021-07-05
Added "Bind to grammar/schema" to command palette

### LINKED ISSUE #514: "Bind to grammar/schema..." should be a command available from the palette
Currently, "Bind to grammar/schema..." can only be triggered if `xml.codeLens.enabled`is set to `true`. 

Screenshot 2021-06-17 at 16 17 17

It makes sense to also expose it as a command available from the command palette

AlexXuChen:
Added new command, "Bind to grammar/schema", to command palette menu.

Closes #514 

### DIFF
--- a/package.json
+++ b/package.json
@@ -524,6 +524,11 @@
         "command": "xml.validation.all.files",
         "title": "Revalidate all opened XML files",
         "category": "XML"
+      },
+      {
+        "command": "xml.command.bind.grammar",
+        "title": "Bind to grammar/schema file",
+        "category": "XML"
       }
     ],
     "menus": {
@@ -531,6 +536,10 @@
         {
           "command": "xml.validation.current.file",
           "when": "editorLangId == xml"
+        },
+        {
+          "command": "xml.command.bind.grammar",
+          "when": "resourceFilename =~ /xml/ && editorIsOpen"
         }
       ]
     },
--- a/src/commands/commandConstants.ts
+++ b/src/commands/commandConstants.ts
@@ -59,6 +59,11 @@ export namespace ClientCommandConstants {
   export const OPEN_BINDING_WIZARD = 'xml.open.binding.wizard';
 
   /**
+   * VSCode client command to open the grammar/schema binding wizard from command menu.
+   */
+  export const COMMAND_PALETTE_BINDING_WIZARD = 'xml.command.bind.grammar';
+
+  /**
    * Client command to execute an XML command on XML Language Server side.
    */
   export const EXECUTE_WORKSPACE_COMMAND = 'xml.workspace.executeCommand';
@@ -85,4 +90,9 @@ export namespace ServerCommandConstants {
    * Command to associate a grammar in a XML document
    */
   export const ASSOCIATE_GRAMMAR_INSERT = "xml.associate.grammar.insert";
+
+  /**
+   * Command to check if the current XML document is bound to a grammar
+   */
+  export const CHECK_BOUND_GRAMMAR = "xml.check.bound.grammar"
 }
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import { commands, ExtensionContext, OpenDialogOptions, Position, QuickPickItem, Uri, window, workspace, WorkspaceEdit } from "vscode";
+import { commands, ExtensionContext, OpenDialogOptions, Position, QuickPickItem, Uri, window, workspace, WorkspaceEdit, Disposable } from "vscode";
 import { CancellationToken, ExecuteCommandParams, ExecuteCommandRequest, ReferencesRequest, TextDocumentIdentifier, TextDocumentEdit } from "vscode-languageclient";
 import { LanguageClient } from 'vscode-languageclient/node';
 import { markdownPreviewProvider } from "../markdownPreviewProvider";
@@ -25,7 +25,7 @@ export async function registerClientServerCommands(context: ExtensionContext, la
 
   registerCodeLensReferencesCommands(context, languageClient);
   registerValidationCommands(context);
-  registerCodeLensAssociationCommands(context, languageClient);
+  registerAssociationCommands(context, languageClient);
 
   // Register client command to execute custom XML Language Server command
   context.subscriptions.push(commands.registerCommand(ClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, (command, ...rest) => {
@@ -139,49 +139,94 @@ for (const label of bindingTypes.keys()) {
 }
 
 /**
- * Register commands used for associating grammar file (XSD,DTD) to a given XML file
+ * The function passed to context subscriptions for grammar association
+ *
+ * @param uri the uri of the XML file path
+ * @param languageClient the language server client
+ */
+async function grammarAssociationCommand (documentURI: Uri, languageClient: LanguageClient) {
+  // A click on Bind to grammar/schema... has been processed in the XML document which is not bound to a grammar
+
+  // Step 1 : open a combo to select the binding type ("standard", "xml-model")
+  const pickedBindingTypeOption = await window.showQuickPick(bindingTypeOptions, { placeHolder: "Binding type" });
+  if(!pickedBindingTypeOption) {
+    return;
+  }
+  const bindingType = bindingTypes.get(pickedBindingTypeOption.label);
+
+  // Open a dialog to select the XSD, DTD to bind.
+  const options: OpenDialogOptions = {
+    canSelectMany: false,
+    openLabel: 'Select XSD or DTD file',
+    filters: {
+      'Grammar files': ['xsd', 'dtd']
+    }
+  };
+
+  const fileUri = await window.showOpenDialog(options);
+  if (fileUri && fileUri[0]) {
+    // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
+    const identifier = TextDocumentIdentifier.create(documentURI.toString());
+    const grammarURI = fileUri[0];
+    try {
+      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
+      // Insert the proper syntax for binding
+      const lspTextDocumentEdit = <TextDocumentEdit>result;
+      const workEdits = new WorkspaceEdit();
+      for (const edit of lspTextDocumentEdit.edits) {
+        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      }
+      workspace.applyEdit(workEdits); // apply the edits
+
+    } catch (error) {
+      window.showErrorMessage('Error during grammar binding: ' + error.message);
+    };
+  }
+}
+
+/**
+ * Register commands used for associating grammar file (XSD,DTD) to a given XML file for command menu and CodeLens
  *
  * @param context the extension context
+ * @param languageClient the language server client
  */
-function registerCodeLensAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
+ function registerAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
+  // For CodeLens
   context.subscriptions.push(commands.registerCommand(ClientCommandConstants.OPEN_BINDING_WIZARD, async (uriString: string) => {
-    // A click on Bind to grammar/schema... has been processed in the XML document which is not bound to a grammar
-    const documentURI = Uri.parse(uriString);
-
-    // Step 1 : open a combo to select the binding type ("standard", "xml-model")
-    const pickedBindingTypeOption = await window.showQuickPick(bindingTypeOptions, { placeHolder: "Binding type" });
-    if(!pickedBindingTypeOption) {
-      return;
-    }
-    const bindingType = bindingTypes.get(pickedBindingTypeOption.label);
-
-    // Open a dialog to select the XSD, DTD to bind.
-    const options: OpenDialogOptions = {
-      canSelectMany: false,
-      openLabel: 'Select XSD or DTD file',
-      filters: {
-        'Grammar files': ['xsd', 'dtd']
-      }
-    };
-
-    const fileUri = await window.showOpenDialog(options);
-    if (fileUri && fileUri[0]) {
-      // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
-      const identifier = TextDocumentIdentifier.create(documentURI.toString());
-      const grammarURI = fileUri[0];
-      try {
-        const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-        // Insert the proper syntax for binding
-        const lspTextDocumentEdit = <TextDocumentEdit>result;
-        const workEdits = new WorkspaceEdit();
-        for (const edit of lspTextDocumentEdit.edits) {
-          workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
-        }
-        workspace.applyEdit(workEdits); // apply the edits
-      } catch (error) {
-        window.showErrorMessage('Error during grammar binding: ' + error.message);
-      };
+    const uri = Uri.parse(uriString);
+    grammarAssociationCommand(uri, languageClient)
+  }));
+  // For command menu
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.COMMAND_PALETTE_BINDING_WIZARD, async () => {
+    const uri = window.activeTextEditor.document.uri;
+    // Run check to ensure available grammar binding command should be executed, or if error is thrown
+    const canBind = await checkCanBindGrammar(uri);
+    if (canBind) {
+      grammarAssociationCommand(uri, languageClient)
+    } else {
+      window.showErrorMessage(`The document ${uri.toString()} is already bound with a grammar`);
     }
   }));
 
-}
+ }
+
+  /**
+   * Change value of 'canBindGrammar' to determine if grammar/schema can be bound
+   *
+   * @param document the text document
+   * @returns the `hasGrammar` check result from server
+   */
+  async function checkCanBindGrammar(documentURI: Uri) {
+        // Retrieve the document uri and identifier
+        const identifier = TextDocumentIdentifier.create(documentURI.toString());
+
+        // Set the custom condition to watch if file already has bound grammar
+        var result = false;
+        try {
+          result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
+        } catch(error) {
+          console.log(`Error while checking bound grammar : ${error}`);
+        }
+
+        return result
+  }


datho7561:
Something that I ran into is that this will fail if the document has no root element. The code for the CodeLens assumes that there is a root element, since the CodeLens is placed right above the root element.

eg. try binding the following file to a grammar:

```xml
```

This will involve adding a check in LemMinX to see if there is a root element when trying to build the edit to add the grammar association.

package.json
```diff
@@ -524,6 +524,11 @@
         "command": "xml.validation.all.files",
         "title": "Revalidate all opened XML files",
         "category": "XML"
+      },
+      {
+        "command": "xml.command.bind.grammar",
+        "title": "Bind to grammar/schema file",
+        "category": "XML"
```
datho7561, line 531: Hide the command when there are no files opened or the current file is not an XML file. (see https://code.visualstudio.com/api/references/when-clause-contexts for a reference on how to do this)

src/commands/commandConstants.ts
```diff
@@ -62,6 +62,8 @@ export namespace ClientCommandConstants {
    * Client command to execute an XML command on XML Language Server side.
    */
   export const EXECUTE_WORKSPACE_COMMAND = 'xml.workspace.executeCommand';
+
+  export const COMMAND_PALETTE_BINDING_WIZARD = 'xml.command.bind.grammar';
```
datho7561, line 66: Put this beside the `OPEN_BINDING_WIZARD` constant declaration, since they are related.

src/commands/registerCommands.ts
```diff
@@ -139,49 +140,76 @@ for (const label of bindingTypes.keys()) {
...
 
-}
\ No newline at end of file
+/**
+ * Register commands used for associating grammar file (XSD,DTD) to a given XML file
+ *
+ * @param context the extension context
+ * @param languageClient the language server client
+ */
+ function registerAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
```
datho7561, line 206: I think it would be a good idea to combine the registration of this and the code lens version into one function, since they are related.

datho7561 (changes requested):

src/commands/registerCommands.ts
AlexXuChen: Will this be affected by one having the uri passed into the function? (i.e. the codelens one receives the uri properly, while my addition doesn't)

datho7561: Nope, its just combining the registration of the commands, not the commands themselves.

src/client/canBindGrammar.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+'use strict';
+
+import { window, workspace, Disposable, TextDocumentContentChangeEvent, TextDocument, Position, commands } from 'vscode';
+
+export function activateCanBindGrammar(bindingWizardProvider: (document: TextDocument, position: Position) => Thenable<boolean>, supportedLanguages: { [id: string]: boolean }): Disposable {
+
+  let disposables: Disposable[] = [];
+  workspace.onDidChangeTextDocument(event => onDidChangeTextDocument(event.document, event.contentChanges), null, disposables);
```
angelozerr, line 19: I think commands.executeCommand('setContext', 'canBindGrammar', result) must be done only when editor has focus and not when editor content changed

You should remove workspace.onDidChangeTextDocument listener  and compute canbind in 
window.onDidChangeActiveTextEditor(computeCanBindingWithGrammar; null, disposables);

angelozerr (changes requested):

src/client/canBindGrammar.ts
datho7561: What if the user manually types in the code to bind to a grammar?

angelozerr: You are right but i find its shame to consume on each changes the canbind lemminx command for this usecase. Perhaps de could skipthis usecase and improvise associategrammarcommand on lemminx side by throwing an exception when dom document is already found with a grammar

src/commands/commandConstants.ts
```diff
@@ -85,4 +90,9 @@ export namespace ServerCommandConstants {
    * Command to associate a grammar in a XML document
    */
   export const ASSOCIATE_GRAMMAR_INSERT = "xml.associate.grammar.insert";
+
+  /**
+   * Command to check if the current XML document is bound to a grammar
+   */
+  export const CHECK_BOUND_GRAMMAR = "xml.check.bound.grammar"
```
angelozerr, line 97: I would rename to 

`export const ASSOCIATE_GRAMMAR_CHECK = "xml.associate.grammar.check";`

src/client/canBindGrammar.ts
```diff
@@ -0,0 +1,65 @@
...
+    isEnabled = false;
+    let editor = window.activeTextEditor;
+    if (!editor) {
+      return;
+    }
+    let document = editor.document;
+    if (!supportedLanguages[document.languageId]) {
+      return;
+    }
+    isEnabled = true;
```
angelozerr, line 37:
```ts
let documentURI= document.uri;
const identifier = TextDocumentIdentifier.create(documentURI.toString());
const result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
commands.executeCommand('setContext', 'canBindGrammar', result)
```

src/client/canBindGrammar.ts
```diff
@@ -0,0 +1,65 @@
...
+  let disposables: Disposable[] = [];
+  workspace.onDidChangeTextDocument(event => onDidChangeTextDocument(event.document, event.contentChanges), null, disposables);
+
+  let isEnabled = false;
+  updateEnabledState();
+  window.onDidChangeActiveTextEditor(updateEnabledState, null, disposables);
+
+  let timeout: NodeJS.Timer | undefined = undefined;
+
+  function updateEnabledState() {
```
angelozerr, line 27: add `async function`

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+import { window, workspace, Disposable, TextDocumentContentChangeEvent, TextDocument, commands } from 'vscode';
+import { TextDocumentIdentifier } from 'vscode-languageclient';
+import { ServerCommandConstants } from './commandConstants';
+
+export function activateCanBindGrammar(supportedLanguages: { [id: string]: boolean }): Disposable {
+
+  let disposables: Disposable[] = [];
+  workspace.onDidChangeTextDocument(event => onDidChangeTextDocument(event.document, event.contentChanges), null, disposables);
+
+  let isEnabled = false;
```
angelozerr, line 23: remove this field

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+  updateEnabledState();
+  window.onDidChangeActiveTextEditor(updateEnabledState, null, disposables);
+
+  let timeout: NodeJS.Timer | undefined = undefined;
+
+  async function updateEnabledState() {
```
angelozerr, line 29: Rename function `updateCanBindGrammar`

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+    let editor = window.activeTextEditor;
```
angelozerr, line 30: Add comments

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (!editor) {
+      return;
+    }
+    let document = editor.document;
+    if (!supportedLanguages[document.languageId]) {
```
angelozerr, line 35: Add comments

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    const documentURI= document.uri;
```
angelozerr, line 39: Add comments

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+    const identifier = TextDocumentIdentifier.create(documentURI.toString());
+    const result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
+    commands.executeCommand('setContext', 'canBindGrammar', result)
+  }
+
+  function onDidChangeTextDocument(document: TextDocument, changes: ReadonlyArray<TextDocumentContentChangeEvent>) {
```
angelozerr, line 45: Remove this listener

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+    let document = editor.document;
+    if (!supportedLanguages[document.languageId]) {
+      return;
+    }
+
+    const documentURI= document.uri;
+    const identifier = TextDocumentIdentifier.create(documentURI.toString());
+    const result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
```
angelozerr, line 41: Try cacth this command call to prevent error from LemMinx side.

src/commands/canBindGrammar.ts
AlexXuChen: Without this listener, there's no way for the extension to know that the file is now bound is there?

src/commands/canBindGrammar.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    let document = editor.document;
+    if (!supportedLanguages[document.languageId]) {
+      return;
+    }
+
+    await checkCanBindGrammar(document);
+  }
+
+  async function onDidChangeTextDocument(document: TextDocument) {
```
angelozerr, line 41: Remove this listener

src/commands/canBindGrammar.ts
AlexXuChen: For example, if I manually remove the binding in the text editor

AlexXuChen:
FOR REFERENCE:

this fix won't solve if the user manually deletes the schema binding, this can be added with an onDidChangeTextDocument listener:

```ts
 async function onDidChangeTextDocument(document: TextDocument) {
    if (typeof timeout !== 'undefined') {
      clearTimeout(timeout);
    }

    await checkCanBindGrammar(document);
  }
  return Disposable.from(...disposables);
```

and called as:
```ts
workspace.onDidChangeTextDocument(event => onDidChangeTextDocument(event.document), null, disposables);
```

angelozerr:
If associategrammarcommand throwing an exception when xml is already bound you dont need to have 
onDidChangeTextDocument(document

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,125 @@ for (const label of bindingTypes.keys()) {
...
+  async function checkCanBindGrammar(document: TextDocument) {
+        // Retrieve the document uri and identifier
+        const documentURI = document.uri;
+        const identifier = TextDocumentIdentifier.create(documentURI.toString());
+
+        // Set the custom condition to watch if file already has bound grammar
+        const result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
+        await commands.executeCommand('setContext', 'canBindGrammar', result);
+
+        return result
```
angelozerr, line 262: Remove thisline code, No need to return result.

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,125 @@ for (const label of bindingTypes.keys()) {
...
+   * @param document the text document
+   * @returns the `hasGrammar` check result from server
+   */
+  async function checkCanBindGrammar(document: TextDocument) {
+        // Retrieve the document uri and identifier
+        const documentURI = document.uri;
+        const identifier = TextDocumentIdentifier.create(documentURI.toString());
+
+        // Set the custom condition to watch if file already has bound grammar
+        const result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
```
angelozerr, line 259: Try catch consume of check bound grammar command

```ts
try {
const result = await commands.executeCommand(ServerCommandConstants.CHECK_BOUND_GRAMMAR, identifier);
        await commands.executeCommand('setContext', 'canBindGrammar', result);
}
catch(error) {
    console.log(`Error while checking bound grammar : ${error}`);
}
```

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,125 @@ for (const label of bindingTypes.keys()) {
...
+    return Disposable.from(...disposables);
+  }
+
+  /**
+   * Change value of 'canBindGrammar' to determine if grammar/schema can be bound
+   *
+   * @param document the text document
+   * @returns the `hasGrammar` check result from server
+   */
+  async function checkCanBindGrammar(document: TextDocument) {
```
angelozerr, line 253: `async function checkCanBindGrammar(documentURI: string) {`

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,125 @@ for (const label of bindingTypes.keys()) {
...
+      if (!editor) {
+        return;
+      }
+      // Check if the document language is supported
+      let document = editor.document;
+      if (!supportedLanguages[document.languageId]) {
+        return;
+      }
+
+      await checkCanBindGrammar(document);
```
angelozerr, line 241: `await checkCanBindGrammar(document.uri);`

After that you canremove TextDocument import.

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,125 @@ for (const label of bindingTypes.keys()) {
...
+   * Perform a check to see if the current (XML) document has a grammar/schema bound to it on text editor open
+   *
+   * @param supportedLanguages the languages the listener should be active for
+   */
+  function activateCanBindGrammar(supportedLanguages: { [id: string]: boolean }): Disposable {
+
+    let disposables: Disposable[] = [];
+
+    // Call listener on text editor open to check if opened (XML) document had a bound grammar/schema
+    updateEnabledState();
```
angelozerr, line 226: updateCheckCanBindGrammar();

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,125 @@ for (const label of bindingTypes.keys()) {
...
+    window.onDidChangeActiveTextEditor(updateEnabledState, null, disposables);
+
+    async function updateEnabledState() {
```
angelozerr, line 229: `async function updateCheckCanBindGrammar();`

src/commands/registerCommands.ts
AlexXuChen: The result check is to see if I can bind the grammar (in cases that the exception isn't thrown) See `registerCommand.ts` line 207

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,129 @@ for (const label of bindingTypes.keys()) {
...
+   *
+   * @param document the text document
+   * @returns the `hasGrammar` check result from server
+   */
+  async function checkCanBindGrammar(documentURI: Uri) {
+        // Retrieve the document uri and identifier
+        const identifier = TextDocumentIdentifier.create(documentURI.toString());
+
+        // Set the custom condition to watch if file already has bound grammar
+        let result;
```
angelozerr, line 258:
```ts
let result = false;
```

src/commands/registerCommands.ts
angelozerr:
```ts
return result;
```

angelozerr: > The result check is to see if I can bind the grammar (in cases that the exception isn't thrown) See registerCommand.ts line 207

Ok thanks, I understand.

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,129 @@ for (const label of bindingTypes.keys()) {
...
+ * @param uriString the string representing the XML file path
+ * @param languageClient the language server client
  */
-function registerCodeLensAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
-  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.OPEN_BINDING_WIZARD, async (uriString: string) => {
-    // A click on Bind to grammar/schema... has been processed in the XML document which is not bound to a grammar
-    const documentURI = Uri.parse(uriString);
+async function grammarAssociationCommand (uriString: string, languageClient: LanguageClient) {
+  // A click on Bind to grammar/schema... has been processed in the XML document which is not bound to a grammar
+  const documentURI = Uri.parse(uriString);
```
angelozerr, line 149: Remove this const declaration and replace method with:

```ts
async function grammarAssociationCommand (documentURI : Uri, languageClient: LanguageClient) {
```

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,129 @@ for (const label of bindingTypes.keys()) {
...
+/**
+ * Register commands used for associating grammar file (XSD,DTD) to a given XML file for command menu and CodeLens
+ *
+ * @param context the extension context
+ * @param languageClient the language server client
+ */
+ function registerAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
+  // For CodeLens
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.OPEN_BINDING_WIZARD, async (uriString: string) => {
+    grammarAssociationCommand(uriString, languageClient)
```
angelozerr, line 200:
```ts
const documentURI = Uri.parse(uriString);
grammarAssociationCommand(documentURI, languageClient);
```

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,129 @@ for (const label of bindingTypes.keys()) {
...
+  }));
+  // For command menu
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.COMMAND_PALETTE_BINDING_WIZARD, async () => {
```
angelozerr, line 203: Replace with:

  ```ts
context.subscriptions.push(commands.registerCommand(ClientCommandConstants.COMMAND_PALETTE_BINDING_WIZARD, async () => {
    const uri = window.activeTextEditor.document.uri;
    // Run check to ensure available grammar binding command should be executed, or if error is thrown
    const canBind = await checkCanBindGrammar(uri);
    if (canBind) {
      grammarAssociationCommand(uri, languageClient)
    }
  }));
```

**Comment by angelozerr** [comment]
After playing with this PR, I think hide command with setContext is not a good idea because:

 * we need to track the binding declaration inside XML file and I find code is complex just to hide a menu.
 * I have in my mind a new feature to give the capability to rebind a XML with an another XSD/DTD file for instance. In this case the command should be visible every time. But it's out the scope of the PR.

In other words, I think should remove the canBindGrammar from package.json and all code which update setContext. When user will try to bind the XML with a grammar and this XML i salready bound, we should just display an error message like I suggested in https://github.com/redhat-developer/vscode-xml/pull/544/files#r673821683

We could improve that in future release by rebind teh XMLwith a new XSD/DTD by improving AssociateGrammarCommand.

File: src/commands/registerCommands.ts (hunk 24 above)
Comment (AlexXuChen) [reply src/commands/registerCommands.ts:200]: This URIString is passed into the registered command as a parameter, as you originally did when you added this command. This is a wrapper, so I believe if I change this, it will change your implementation of the association command.

File: src/commands/registerCommands.ts [hunk 26]
```diff
@@ -139,49 +139,97 @@ for (const label of bindingTypes.keys()) {
...
+      // Insert the proper syntax for binding
+      const lspTextDocumentEdit = <TextDocumentEdit>result;
+      const workEdits = new WorkspaceEdit();
+      for (const edit of lspTextDocumentEdit.edits) {
+        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
       }
+      workspace.applyEdit(workEdits); // apply the edits
+
+      // Hide the "Bind to grammar/schema" command
+      commands.executeCommand('setContext', 'canBindGrammar', false)
```
angelozerr, line 183: remove this line code

src/commands/registerCommands.ts
angelozerr: if you see https://github.com/redhat-developer/vscode-xml/blob/a720559fae8f8dd93c14a152dc3a8aa066e9732b/src/commands/registerCommands.ts#L149 the follwing code use an Uri instance and nor a string. So grammarAssociationCommand can have a Uri instanceisntead of String.

In case with Codelens, we convert uri string to an Uri before calling grammarAssociationCommand . In the case of command palette we have laready Uri instance from textDocument, so grammarAssociationCommand  can have an Uri instance instead of string.

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,93 @@ for (const label of bindingTypes.keys()) {
...
+ function registerAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
+  // For CodeLens
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.OPEN_BINDING_WIZARD, async (uri: Uri) => {
+    grammarAssociationCommand(uri, languageClient)
+  }));
+  // For command menu
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.COMMAND_PALETTE_BINDING_WIZARD, async () => {
+    const uri = window.activeTextEditor.document.uri;
+    // Run check to ensure available grammar binding command should be executed, or if error is thrown
+    const canBind = await checkCanBindGrammar(window.activeTextEditor.document.uri);
```
angelozerr, line 202: Use uri

src/commands/registerCommands.ts
```diff
@@ -139,49 +139,93 @@ for (const label of bindingTypes.keys()) {
...
+/**
+ * Register commands used for associating grammar file (XSD,DTD) to a given XML file for command menu and CodeLens
+ *
+ * @param context the extension context
+ * @param languageClient the language server client
+ */
+ function registerAssociationCommands(context: ExtensionContext, languageClient: LanguageClient) {
+  // For CodeLens
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.OPEN_BINDING_WIZARD, async (uri: Uri) => {
+    grammarAssociationCommand(uri, languageClient)
```
angelozerr, line 196: The parameter uri is a string for codelens. Déclare parameter with uriString:string and convertir it 
const uri= Uri.parse(uriString);

angelozerr (changes requested):

angelozerr:
Code is now perfect and I play with the lemMinx / vscode-xml PR and it works great. There is some usecase that we must improve but we can do that in an another PR (ex : try to bind with XML which have no root tag element, bind a xml-model, etc.).

I will create an issue for that as soon as this PR are merged. Once LemMinx PR will be merged (see my comments), I'm OK to merge this PR.

angelozerr (approved):

datho7561 (approved):

PR #545, opened 2021-07-07
Made resolveExternalEntities depend of workspace trust

### LINKED ISSUE #537: have an option to make "resolve external entities" depending on workspace trust
Ideally I'd like to have the setting enabled in trusted workspaces, but be disabled in untrusted workspaces.
Would it be reasonable to switch the current setting from boolean (always/trusted/never) to an enum or should this be a new setting?

AlexXuChen:
Changed the resolveExternalEntities setting to depend on workspace trust, where it uses user settings in trusted workspaces, and is false in untrusted workspaces.

Closes #537 

### DIFF
--- a/src/settings/settings.ts
+++ b/src/settings/settings.ts
@@ -222,6 +222,14 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
     enabled: workspace.getConfiguration('redhat.telemetry').get('enabled', false)
   };
 
+  // Check workspace trust
+  const isWorkspaceTrusted = (workspace as any).isTrusted;
+  if (isWorkspaceTrusted !== undefined && !isWorkspaceTrusted) {
+    xml['xml']['validation']['resolveExternalEntities'] = false;
+  } else {
+    xml['xml']['validation']['resolveExternalEntities'] = workspace.getConfiguration('xml').get('validation.resolveExternalEntities', false);
+  }
+
   //applying externalXmlSettings to the xmlSettings
   externalXmlSettings.xmlCatalogs.forEach(catalog => {
     if (!xml['xml']['catalogs'].includes(catalog)) {


src/settings/settings.ts
```diff
@@ -215,6 +215,7 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
   xml['xml']['logs']['file'] = logfile;
   xml['xml']['useCache'] = true;
   xml['xml']['java']['home'] = javaHome;
+  xml['xml']['validation']['resolveExternalEntities'] = workspace.getConfiguration('files').get('resolveExternalEntities', false);
```
datho7561, line 218: This reads `files.resolveExternalEntities` from the user's settings. You need to read `xml.validation.resolveExternalEntities` from the user's settings. As well, you need to set this setting to `false` if the user doesn't trust the workspace.

src/settings/settings.ts
```diff
@@ -222,6 +222,14 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
     enabled: workspace.getConfiguration('redhat.telemetry').get('enabled', false)
   };
 
+  // Check workspace trust
+  const isWorkspaceTrusted = (workspace as any).isTrusted;
+  if (isWorkspaceTrusted) {
```
datho7561, line 227: If workspace trust isn't implemented in the client (eg. Theia), this works properly.
However, if workspace trust is implemented and the workspace is trusted, resolveExternalEntities will be disabled, which is the opposite of what we want.
I think if you explicitly check if `isTrusted` is `undefined` or if it is `false`, then you can avoid this issue.

datho7561 (changes requested):

src/settings/settings.ts
AlexXuChen: Oh sorry, there is supposed to be a ! infront

src/settings/settings.ts
```diff
@@ -222,6 +222,14 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
     enabled: workspace.getConfiguration('redhat.telemetry').get('enabled', false)
   };
 
+  // Check workspace trust
+  const isWorkspaceTrusted = (workspace as any).isTrusted;
+  if (!isWorkspaceTrusted) {
```
datho7561, line 227: If `workspace.isTrusted` does not exist in the API, i.e. in Theia, `isWorkspaceTrusted` becomes `undefined`. This makes it impossible to enable `resolveExternalEntities` in this case.

datho7561 (changes requested):

src/settings/settings.ts
AlexXuChen: The ! doesnt solve this? I should say it explicitly like you mentioned then?

datho7561: Yes, since `!undefined` is true and `!false` is also true

datho7561 (approved):

GitMensch:
The change is good, but I think this really needs to be documented, at least on the setting itself.

AlexXuChen:
Documentation: #546

PR #548, opened 2021-07-13
Handle OOM in Java language server

### LINKED ISSUE #527: XML Language Server keeps restarting when running out of memory
Telemetry data tells us a few users (< 0.1%) have their XML Language Server restarting every minute. Since this is a slower rate than 5 times in less than 3 minutes, it doesn't bubble up to the user. 
Still, this indicates a problem and I have no idea what can cause it. Until we find a solution, we need to get better information (from telemetry) to figure out what's causing this.

datho7561:
Add parameters to the java server to make lemminx crash when it runs out of memory. Detect when the java server shuts down due to running out of memory, display a message to the user that this happened, and don't attempt to restart the server.

Closes #527


datho7561:
@fbricon it seems like it isn't possible to do anything on the lemminx side when we hit an OOM Error. By crashing lemminx and checking for the heap dump, we can notify the user that its an OOM and prevent the server from restarting.

docs/Troubleshooting.md
```diff
@@ -21,3 +21,23 @@ You can kill the process by:
...
+
+#### How to increase the amount of memory available to the Language Server
+
+If you are using the Java Language Server:
+1. Go to settings
+2. Navigate to the setting `xml.server.vmargs`
+3. Add `-Xmx 1G` to the setting string. This allows the the language server to use at most 1 gigabyte of memory.
+4. If the problem persists, you can increase the `1G` to `2G` or higher
+
+If you are using the binary language server:
```
rgrunber, line 41: I probably wouldn't format the binary language server part as "steps" since this is more informational. I would maybe mention that OOM issues are less likely to occur with the binary language server. Maybe we could mention that those running into memory issues should be encouraged to try the binary server as a workaround ?

rgrunber (changes requested):
This worked pretty well for me. I think it's a good first step. We might be able to improve things if the language server client had a nice way to access the exit code of the LS executable, but not possible for now it seems.

src/client/clientErrorHandler.ts
```diff
@@ -23,6 +28,11 @@ export class ClientErrorHandler implements ErrorHandler {
 
   closed(): CloseAction {
     this.restarts.push(Date.now());
+    const heapProfileGlob = new glob.GlobSync(`${path.resolve(__dirname, '..')}/java_*.hprof`);
+    if (heapProfileGlob.found.length) {
+      showOOMMessage();
```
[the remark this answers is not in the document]
datho7561: I was originally thinking that the user could read the heap dump files to help find memory leaks. However, since the files are a binary format, I think its a good idea to clean them up here.

src/server/java/javaServerStarter.ts
```diff
@@ -63,6 +78,18 @@ function prepareParams(requirements: RequirementsData, xmlJavaExtensions: string
...
     }
   }
+  if (vmargs.indexOf(CRASH_ON_OOM) < 0) {
+    params.push(CRASH_ON_OOM);
+  }
+  if (vmargs.indexOf(HEAP_DUMP) < 0) {
+    params.push(HEAP_DUMP);
+  }
+  if (vmargs.indexOf(HEAP_DUMP_LOCATION) < 0) {
+    params.push(`${HEAP_DUMP_LOCATION}${path.resolve(__dirname, '..')}`);
```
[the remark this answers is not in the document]
datho7561: It resolves to the directly to the folder you mentioned. I'll look into workspace storage, I have never used it before.

docs/Troubleshooting.md
```diff
@@ -21,3 +21,22 @@ You can kill the process by:
...
+
+If you get an Out of Memory Error, but aren't working with large XML files,
+then there may be a memory leak in the language server.
+Please [file a issue](https://github.com/redhat-developer/vscode-xml/issues) with a description of what you were doing if this is the case.
+
+#### How to increase the amount of memory available to the Java Language Server
+
+1. Go to settings
+2. Navigate to the setting `xml.server.vmargs`
+3. Add `-Xmx 1G` to the setting string. This allows the the language server to use at most 1 gigabyte of memory.
```
rgrunber, line 41: -Xmx1G is the default. Should we maybe set this to 1.5G ?

rgrunber (commented):
Looks good. Just a minor detail to change.

Just wondering (whether it's possible, and worth it), if a user does set `-XX:HeapDumpPath=..`, could we listen to that location for newly added files ?

rgrunber (approved):

docs/Troubleshooting.md
```diff
@@ -21,3 +21,22 @@ You can kill the process by:
...
+
+If you get an Out of Memory Error, but aren't working with large XML files,
+then there may be a memory leak in the language server.
+Please [file a issue](https://github.com/redhat-developer/vscode-xml/issues) with a description of what you were doing if this is the case.
+
+#### How to increase the amount of memory available to the Java Language Server
+
+1. Go to settings
+2. Navigate to the setting `xml.server.vmargs`
+3. Add `-Xmx 1.5G` to the setting string. This allows the the language server to use at most 1.5 gigabytes of memory.
```
fbricon, line 41: there's no space between -Xmx and the value

-Xmx512m would probably be good enough. We're starting from 64M https://github.com/redhat-developer/vscode-xml/blob/master/package.json#L122. 1.5G is a huge step forward

rgrunber: My mistake. I think I was looking at vscode-java :| . Also, yes, I figured maybe it was fine with the space, but looks like it may not get set this way.

datho7561:
> Just wondering (whether it's possible, and worth it), if a user does set `-XX:HeapDumpPath=..`, could we listen to that location for newly added files ?

I implemented a strategy that listens to the HeapDumpPath specified by the JVM arguments if the user sets it there. Unlike when the argument is set in the typescript code, it won't delete the heap dumps that are generated, since presumably the user wants to read them.

src/utils/contextUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+import { ExtensionContext } from "vscode";
+
+/**
+ * Returns the path to the extension storage given the extension context
+ *
+ * @param context the extension context
+ * @returns the path to the extension storage
+ */
+export function getGlobalStoragePath(context: ExtensionContext): string {
+  let storagePath = context.globalStorageUri.path;
```
rgrunber, line 23: Did you try `fsPath` from https://code.visualstudio.com/api/references/vscode-api#Uri . Docs seem to indicate it should handle what you want.

datho7561: Thank you, this is what I needed!

datho7561:
@fbricon does this PR work for you? You should see a popup when the language server crashes due to an OOM Error

src/server/java/jvmArguments.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ *  Copyright (c) 2021 Red Hat, Inc. and others.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-v20.html
+ *
+ *  Contributors:
+ *  Red Hat Inc. - initial API and implementation
+ *  Microsoft Corporation - Auto Closing Tags
```
fbricon, line 10: auto closing tags???

src/server/java/javaServerStarter.ts
```diff
@@ -1,16 +1,27 @@
...
 
+/**
+ * Argument that tells the program to crash when an OutOfMemoryError is raised
+ */
+const CRASH_ON_OOM = '-XX:+ExitOnOutOfMemoryError';
+
+/**
+ * Argument that tells the program to generate a heap dump file when an OutOfMemoryError is raised
+ */
+const HEAP_DUMP = '-XX:+HeapDumpOnOutOfMemoryError';
```
fbricon, line 23: should move to jvmArguments.ts for consistency

src/server/java/javaServerStarter.ts
```diff
@@ -1,16 +1,27 @@
...
+import glob = require('glob');
 
 declare var v8debug;
 
 const DEBUG = (typeof v8debug === 'object') || startedInDebugMode();
 
+/**
+ * Argument that tells the program to crash when an OutOfMemoryError is raised
+ */
+const CRASH_ON_OOM = '-XX:+ExitOnOutOfMemoryError';
```
fbricon, line 18: should move to jvmArguments.ts for consistency

src/client/clientErrorHandler.ts
```diff
@@ -23,11 +32,22 @@ export class ClientErrorHandler implements ErrorHandler {
...
+      cleanUpHeapDumps(this.context);
+      showOOMMessage();
+      return CloseAction.DoNotRestart;
+    }
     if (this.restarts.length < 5) {
       return CloseAction.Restart;
     } else {
       const diff = this.restarts[this.restarts.length - 1] - this.restarts[0];
-      if (diff <= 3 * 60 * 1000) {
+      if (diff <= 5 * 60 * 1000) {
```
rgrunber, line 50: The error message just below as well. But actually, do you have an opposition to making it 10 minutes ? Basically, 1 restart every 2 minutes or less (repeated 5 times) would trigger it.

datho7561: Okay. I have set it to 5 crashes in the last 10 minutes

src/client/clientErrorHandler.ts
```diff
@@ -23,11 +33,23 @@ export class ClientErrorHandler implements ErrorHandler {
...
+      showOOMMessage();
+      return CloseAction.DoNotRestart;
+    }
     if (this.restarts.length < 5) {
       return CloseAction.Restart;
     } else {
       const diff = this.restarts[this.restarts.length - 1] - this.restarts[0];
-      if (diff <= 3 * 60 * 1000) {
+      if (diff <= 10 * 60 * 1000) {
         window.showErrorMessage(`The ${this.name} language server crashed 5 times in the last 3 minutes. The server will not be restarted.`);
```
rgrunber, line 53: "last 10 minutes"

rgrunber:
I can confirm that data would be broadcast correctly upon crash, and for opening the link. @fbricon , if you're fine with this, I think we can probably merge.

src/client/clientErrorHandler.ts
```diff
@@ -23,12 +33,24 @@ export class ClientErrorHandler implements ErrorHandler {
...
     this.restarts.push(Date.now());
+    const heapProfileGlob = new glob.GlobSync(`${this.heapDumpFolder}/java_*.hprof`);
+    if (heapProfileGlob.found.length) {
+      // Only clean heap dumps that are generated in the default location.
+      // The default location is the extension global storage
+      // This means that if users change the folder where the heap dumps are placed,
+      // then they will be able to read the heap dumps,
+      // since they aren't immediately deleted.
+      cleanUpHeapDumps(this.context);
+      Telemetry.sendTelemetry(Telemetry.JAVA_OOM_EVT);
```
fbricon, line 44: I'd send the Xmx value that triggered the OOME

datho7561: We should be able to collect Xmx from the server startup event, although it might be a good idea to collect it here to catch if they allocate so little RAM that the server can't start.

datho7561: Actually, I think the value from the startup telemetry is better. The user could have changed the value of Xmx in their settings after the server started. The Xmx from server startup telemetry will give the actual Xmx used by the JVM.

fbricon: It's infinitely easier to understand the failure on our backend, if the Xmx value is attached to the error. It'll likely be correct in 99+% of cases

datho7561: Okay. This should be collected in this PR now

rgrunber:
I think we can merge this when ready.

PR #551, opened 2021-07-16
Added `xml.restart.language.server` command for language server connection restart

### LINKED ISSUE #539: add command "Restart Language Server"
... as seen in the Python extension `python.analysis.restartLanguageServer`.

Not a priority at all, but something to possibly consider.

Reasoning: sometimes the server seems to hang and if it doesn't restart its own (which may take some seconds to be triggered, I don't know if there are other cases when this happens apart from "server chrashed") you have three options: a command like this or a manual killing of the process tree (after you've found the correct java process) or (which will loose your undo and position history) the reload window command.

AlexXuChen:
Added `xml.restart.language.server` command to trigger language server connection restart.
Peek 2022-07-20 13-16

Closes #539 

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -1,10 +1,11 @@
 import * as path from 'path';
-import { commands, ExtensionContext, OpenDialogOptions, Position, QuickPickItem, Uri, window, workspace, WorkspaceEdit, Disposable, ConfigurationTarget, TextDocument, WorkspaceFolder } from "vscode";
+import { commands, ExtensionContext, OpenDialogOptions, Position, QuickPickItem, Uri, window, workspace, WorkspaceEdit, ConfigurationTarget, TextDocument } from "vscode";
 import { CancellationToken, ExecuteCommandParams, ExecuteCommandRequest, ReferencesRequest, TextDocumentIdentifier, TextDocumentEdit } from "vscode-languageclient";
 import { LanguageClient } from 'vscode-languageclient/node';
 import { markdownPreviewProvider } from "../markdownPreviewProvider";
 import { ClientCommandConstants, ServerCommandConstants } from "./commandConstants";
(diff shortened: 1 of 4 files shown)


AlexXuChen:
If someone else can quickly confirm for me, when running the command, the client command registration seems to re-trigger, which tells me that the language client is restarting. Can someone confirm this, or give me another check to see if the language client has restarted?

GitMensch:
Thanks for the PR!

The best option to check this is to inspect the processes (`top -u $(whoami)` or Taskmanager/Sysinternals ProcessExplorer on win32). Before and after issuing the command.

GitMensch:
Is there a reason that this is still on draft? The original implementation seemed fine and the linked issue is still open...

src/commands/registerCommands.ts
```diff
@@ -322,4 +323,19 @@ async function checkCanBindGrammar(documentURI: Uri) {
...
+
+/**
+ * Register command to restart the connection to the language server
+ *
+ * @param context the extension context
+ * @param languageClient the language client
+ */
+function registerRestartLanguageServerCommand(context: ExtensionContext, languageClient: LanguageClient) {
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.RESTART_LANGUAGE_SERVER, async () => {
+    // Can be replaced with `await languageClient.restart()` with vscode-languageclient ^8.0.0,
```
AlexXuChen, line 336: Use 
```ts
await languageClient.restart()
```
once we move to `vscode-languageclient ^8.0.1`

rgrunber (approved):
Change works very well for me.

I was considering just leaving the timeout even when run outside of debug mode, but given that restart() will have the timeout, we may as well run it that way.

rgrunber:
@AlexXuChen , may be worth adding to https://github.com/redhat-developer/vscode-xml/blob/master/docs/Commands.md as well.

Issue #562, opened 2021-07-28
Add CodeLens support in XML features docs


angelozerr:
In https://github.com/redhat-developer/vscode-xml/blob/master/docs/Features/XMLFeatures.md add Codelens support and a subsection about Codelens `Bind grammar/schema`

PR #565, opened 2021-07-28
added 'Bind to grammar/schema file' command to Commands.md

### LINKED ISSUE #561: Add `Bind to grammar/schema file` command in docs
In https://github.com/redhat-developer/vscode-xml/blob/master/docs/Commands.md#commands it should be good to add the `Bind to grammar/schema file` command.

AlexXuChen:
Added documentation for 'Bind to grammar/schema file' command in `Commands.md`.

Closes #561


docs/Commands.md
```diff
@@ -4,9 +4,15 @@
 
 ![XML Commands](images/Commands/XMLCommands.png)
 
+## Bind to grammar/schema file
+
+This command triggers the [XML Binding Wizard]() for the current file.
```
AlexXuChen, line 9: I think we should document the XML Binding Wizard feature somewhere (unless I missed it) so I can link it here.
Thoughts @angelozerr?

AlexXuChen: If so, it can also be linked in #562

angelozerr: Indeed, we should have a new `BindingWithGrammar.md` file referenced in https://github.com/redhat-developer/vscode-xml/tree/master/docs#user-guide.


BindingWithGrammar.md could looks like:

## Binding with  existing grammar

In the following sections, we would like to bind the XML `foo.xml` file:

```xml
<foo>
  <bar />
</foo>
```

with the XSD `foo.xsd` file (in the same folder as foo.xml) :

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:element name="foo">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="bar" />
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

### Command

### Codelens

### Quickfix

## Binding with new grammar

Imagine you have the XML `foo.xml` file:

```xml
<foo>
  <bar />
</foo>
```

and you want to generate a grammar and bind it.

### Quickfix

AlexXuChen: Should I open a separate issue and pr for this new page then?

AlexXuChen:
Need to merge #570 before this change

datho7561 (approved):
LGTM

PR #584, opened 2021-08-12
added file association option to binding wizard

### LINKED ISSUE #573: Binding grammar/schema with file associations
Today we manage `Binding grammar/schema `with the binding type:

 * standard 
 * XML association:

image

It should be nice to support the [file associations](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Preferences.md#file-associations) binding type.

When file association will be selected, it should show an input with the glob pattern (prefilled with the xml file name). When wizard will be finished, it should update the `xml.fileAssociations` settings.

AlexXuChen:
Added "File association" option to binding wizard command to bind grammar through `settings.json`

Closes #573 

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -1,9 +1,10 @@
 import * as path from 'path';
-import { commands, ExtensionContext, OpenDialogOptions, Position, QuickPickItem, Uri, window, workspace, WorkspaceEdit, Disposable } from "vscode";
+import { commands, ExtensionContext, OpenDialogOptions, Position, QuickPickItem, Uri, window, workspace, WorkspaceEdit, Disposable, ConfigurationTarget, TextDocument, WorkspaceFolder } from "vscode";
 import { CancellationToken, ExecuteCommandParams, ExecuteCommandRequest, ReferencesRequest, TextDocumentIdentifier, TextDocumentEdit } from "vscode-languageclient";
 import { LanguageClient } from 'vscode-languageclient/node';
 import { markdownPreviewProvider } from "../markdownPreviewProvider";
 import { ClientCommandConstants, ServerCommandConstants } from "./commandConstants";
+import { getRelativePath, getFileName, getDirectoryPath, getWorkspaceUri } from '../utils/fileUtils';
 
 /**
  * Register the commands for vscode-xml that don't require communication with the language server
@@ -143,7 +144,8 @@ function registerValidationCommands(context: ExtensionContext) {
 
 export const bindingTypes = new Map<string, string>([
   ["Standard (xsi, DOCTYPE)", "standard"],
-  ["XML Model association", "xml-model"]
+  ["XML Model association", "xml-model"],
+  ["File association", "fileAssociation"]
 ]);
 
 const bindingTypeOptions: QuickPickItem[] = [];
@@ -154,13 +156,13 @@ for (const label of bindingTypes.keys()) {
 /**
  * The function passed to context subscriptions for grammar association
  *
- * @param uri the uri of the XML file path
+ * @param documentURI the uri of the XML file path
  * @param languageClient the language server client
  */
 async function grammarAssociationCommand(documentURI: Uri, languageClient: LanguageClient) {
   // A click on Bind to grammar/schema... has been processed in the XML document which is not bound to a grammar
 
-  // Step 1 : open a combo to select the binding type ("standard", "xml-model")
+  // Step 1 : open a combo to select the binding type ("standard", "xml-model", "fileAssociation")
   const pickedBindingTypeOption = await window.showQuickPick(bindingTypeOptions, { placeHolder: "Binding type" });
   if (!pickedBindingTypeOption) {
     return;
@@ -179,22 +181,88 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
   const fileUri = await window.showOpenDialog(options);
   if (fileUri && fileUri[0]) {
     // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
-    const identifier = TextDocumentIdentifier.create(documentURI.toString());
     const grammarURI = fileUri[0];
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      const currentFile = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ? window.activeTextEditor.document : undefined;
+      if (bindingType == 'fileAssociation') {
+        // Bind grammar using file association
+        await bindWithFileAssociation(documentURI, grammarURI, currentFile);
+      } else {
+        // Bind grammar using standard binding
+        await bindWithStandard(documentURI, grammarURI, bindingType, languageClient);
       }
-      workspace.applyEdit(workEdits); // apply the edits
-
     } catch (error) {
       window.showErrorMessage('Error during grammar binding: ' + error.message);
     };
   }
+}
+
+/**
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ * @param document the opened TextDocument
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, document: TextDocument) {
+  const absoluteGrammarFilePath = grammarURI.toString();
+  const currentFilename = getFileName(documentURI.toString());
+  const currentWorkspaceUri = getWorkspaceUri(document);
+  // If the grammar file is in the same workspace, use the relative path, otherwise use the absolute path
+  const grammarFilePath = getDirectoryPath(absoluteGrammarFilePath).includes(currentWorkspaceUri.toString()) ? getRelativePath(currentWorkspaceUri.toString(), absoluteGrammarFilePath) : absoluteGrammarFilePath;
+
+  const defaultPattern = `**/${currentFilename}`;
+  const inputBoxOptions = {
+    title: "File Association Pattern",
+    value: defaultPattern,
+    placeHolder: defaultPattern,
+    prompt: "Enter the pattern of the XML document(s) to be bound."
+  }
+  const inputPattern = (await window.showInputBox(inputBoxOptions));
+  if (!inputPattern) {
+    // User closed the input box with Esc
+    return;
+  }
+  const fileAssociation = {
+    "pattern": inputPattern,
+    "systemId": grammarFilePath
+  }
+  addToValueToSettingArray("xml.fileAssociations", fileAssociation);
+}
+
+/**
+ * Bind grammar file using standard XML document grammar
+ *
+ * @param documentURI the URI of the XML file path
+ * @param grammarURI the URI of the user selected grammar file
+ * @param bindingType the selected grammar binding type
+ * @param languageClient the language server client
+ */
+async function bindWithStandard(documentURI: Uri, grammarURI: Uri, bindingType: string, languageClient: LanguageClient) {
+  const identifier = TextDocumentIdentifier.create(documentURI.toString());
+  const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
+  // Insert the proper syntax for binding
+  const lspTextDocumentEdit = <TextDocumentEdit>result;
+  const workEdits = new WorkspaceEdit();
+  for (const edit of lspTextDocumentEdit.edits) {
+    workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+  }
+  workspace.applyEdit(workEdits); // apply the edits
+}
+
+/**
+ * Add an entry, value, to the setting.json field, key
+ *
+ * @param key the filename/path of the xml document
+ * @param value the object to add to the config
+ */
+function addToValueToSettingArray<T>(key: string, value: T): void {
+  const settingArray: T[] = workspace.getConfiguration().get<T[]>(key, []);
+  if (settingArray.includes(value)) {
+    return;
+  }
+  settingArray.push(value);
+  workspace.getConfiguration().update(key, settingArray, ConfigurationTarget.Workspace);
 }
 
 /**
--- a/src/settings/variableSubstitution.ts
+++ b/src/settings/variableSubstitution.ts
@@ -1,5 +1,6 @@
 import * as path from "path";
-import { TextDocument, window, workspace, WorkspaceFolder } from "vscode";
+import { window } from "vscode";
+import { getFilePath, getWorkspaceUri } from "../utils/fileUtils";
 import { XMLFileAssociation } from "../api/xmlExtensionApi";
 
 /**
@@ -99,11 +100,9 @@ const VARIABLE_SUBSTITUTIONS: VariableSubstitution[] = [
 export function getVariableSubstitutedAssociations(associations: XMLFileAssociation[]): XMLFileAssociation[] {
 
   // Collect properties needed to resolve variables
-  const currentFile: TextDocument = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ?  window.activeTextEditor.document : undefined;
-  const currentFileUri: string = (currentFile && currentFile.uri) ? currentFile.uri.fsPath : undefined;
-  const currentWorkspace: WorkspaceFolder = (currentFile && currentFile.uri) ?  workspace.getWorkspaceFolder(currentFile.uri) : undefined;
-  const currentWorkspaceUri: string = (currentWorkspace && currentWorkspace.uri.fsPath)
-    || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath);
+  const currentFile = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ? window.activeTextEditor.document : undefined;
+  const currentFileUri = getFilePath(currentFile);
+  const currentWorkspaceUri = getWorkspaceUri(currentFile);
 
   // Remove variables that can't be resolved
   let variablesToSubstitute = VARIABLE_SUBSTITUTIONS;
@@ -123,7 +122,7 @@ export function getVariableSubstitutedAssociations(associations: XMLFileAssociat
   const subVars = (val: string): string => {
     let newVal = val;
     for (const settingVariable of variablesToSubstitute) {
-      newVal = settingVariable.substituteString(newVal, currentFileUri, currentWorkspaceUri);
+      newVal = settingVariable.substituteString(newVal, currentFileUri, currentWorkspaceUri.toString());
     }
     return newVal;
   }
--- a/src/utils/fileUtils.ts
+++ b/src/utils/fileUtils.ts
@@ -1,0 +1,55 @@
+import * as path from "path";
+import { TextDocument, Uri, workspace, WorkspaceFolder } from "vscode";
+
+/**
+ * Return the current workspace uri from root
+ *
+ * @param document the opened TextDocument
+ * @returns the path from root of the current workspace
+ */
+ export function getWorkspaceUri(document: TextDocument): Uri | undefined {
+  const currentWorkspace: WorkspaceFolder = (document && document.uri) ? workspace.getWorkspaceFolder(document.uri) : undefined;
+  return ((currentWorkspace && currentWorkspace.uri) || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri));
+}
+
+/**
+ * Return the uri of the current file
+ *
+ * @param document the opened TextDocument
+ * @returns the uri of the current file
+ */
+export function getFilePath(document: TextDocument): string {
+  return (document && document.uri) ? document.uri.fsPath : undefined;
+}
+
+/**
+ * Uses path to return a basename from a uri
+ *
+ * @param filePath the uri of the file
+ * @return the filename
+ */
+export function getFileName(filePath: string): string {
+  return path.basename(filePath);
+}
+
+/**
+ * Return the relative file path between a start and destination uri
+ *
+ * @param startPath the absolute path of the beginning directory
+ * @param destinationPath the absolute path of destination file
+ * @returns the path to the destination relative to the start
+ */
+export function getRelativePath(startPath: string, destinationPath: string): string {
+  return path.relative(path.normalize(startPath), path.normalize(destinationPath)).replace(/\\/g, '/');
+}
+
+/**
+ * Uses path to return the directory name from a uri
+ *
+ * @param filePath the uri of the file
+ * @return the directory path
+ */
+export function getDirectoryPath(filePath: string): string {
+  return path.dirname(filePath);
+
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -171,7 +171,7 @@
     },
     "@types/yauzl": {
       "version": "2.9.1",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/@types/yauzl/-/yauzl-2.9.1.tgz",
+      "resolved": "https://registry.npmjs.org/@types/yauzl/-/yauzl-2.9.1.tgz",
       "integrity": "sha512-A1b8SU4D10uoPjwb0lnHmmu8wZhR9d+9o2PKBQT2jU5YPTKsxac6M2qGAdY7VcL+dHHhARVUDmeg0rOrcd9EjA==",
       "dev": true,
       "requires": {
@@ -403,7 +403,7 @@
     },
     "ajv-keywords": {
       "version": "3.5.2",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/ajv-keywords/-/ajv-keywords-3.5.2.tgz",
+      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-3.5.2.tgz",
       "integrity": "sha512-5p6WTN0DdTGVQk6VjcEju19IgaHudalcfabD7yhDGeA6bcQnmL+CpveLJq/3hvfwd1aof6L386Ougkx6RfyMIQ==",
       "dev": true
     },
@@ -869,7 +869,7 @@
     },
     "buffer-crc32": {
       "version": "0.2.13",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
+      "resolved": "https://registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
       "integrity": "sha1-DTM+PwDqxQqhRUq9MO+MKl2ackI="
     },
     "buffer-equal": {
@@ -1720,13 +1720,13 @@
     },
     "fast-deep-equal": {
       "version": "3.1.3",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
+      "resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
       "integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q==",
       "dev": true
     },
     "fast-json-stable-stringify": {
       "version": "2.1.0",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/fast-json-stable-stringify/-/fast-json-stable-stringify-2.1.0.tgz",
+      "resolved": "https://registry.npmjs.org/fast-json-stable-stringify/-/fast-json-stable-stringify-2.1.0.tgz",
       "integrity": "sha512-lhd/wF+Lk98HZoTCtlVraHtfh5XYijIjalXck7saUtuanSDyLMxnHhSXEDJqHxD7msR8D0uCmqlkwjCV8xvwHw==",
       "dev": true
     },
@@ -1744,7 +1744,7 @@
     },
     "fd-slicer": {
       "version": "1.1.0",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/fd-slicer/-/fd-slicer-1.1.0.tgz",
+      "resolved": "https://registry.npmjs.org/fd-slicer/-/fd-slicer-1.1.0.tgz",
       "integrity": "sha1-JcfInLH5B3+IkbvmHY85Dq4lbx4=",
       "requires": {
         "pend": "~1.2.0"
@@ -2975,7 +2975,7 @@
     },
     "mkdirp": {
       "version": "0.5.5",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/mkdirp/-/mkdirp-0.5.5.tgz",
+      "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.5.tgz",
       "integrity": "sha512-NKmAlESf6jMGym1++R0Ra7wvhV+wFW63FaSOFPwRahvea0gMUcGUhVeAg/0BC0wiv9ih5NYPB1Wn1UEI1/L+xQ==",
       "dev": true,
       "requires": {
@@ -3042,7 +3042,7 @@
     },
     "neo-async": {
       "version": "2.6.2",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/neo-async/-/neo-async-2.6.2.tgz",
+      "resolved": "https://registry.npmjs.org/neo-async/-/neo-async-2.6.2.tgz",
       "integrity": "sha512-Yd3UES5mWCSqR+qNT93S3UoYUkqAZ9lLg8a7g9rimsWmYGK8cVToA4/sF3RrshdyV3sAGMXVUmpMYOw+dLpOuw==",
       "dev": true
     },
@@ -3377,7 +3377,7 @@
     },
     "pend": {
       "version": "1.2.0",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/pend/-/pend-1.2.0.tgz",
+      "resolved": "https://registry.npmjs.org/pend/-/pend-1.2.0.tgz",
       "integrity": "sha1-elfrVQpng/kRUzH89GY9XI4AelA="
     },
     "picomatch": {
@@ -3956,7 +3956,7 @@
     },
     "source-map-support": {
       "version": "0.5.19",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/source-map-support/-/source-map-support-0.5.19.tgz",
+      "resolved": "https://registry.npmjs.org/source-map-support/-/source-map-support-0.5.19.tgz",
       "integrity": "sha512-Wonm7zOCIJzBGQdB+thsPar0kYuCIzYvxZwlBa87yi/Mdjv7Tip2cyVbLj5o0cFPN4EVkuTwb3GDDyUx2DGnGw==",
       "dev": true,
       "requires": {
@@ -4980,7 +4980,7 @@
     },
     "yauzl": {
       "version": "2.10.0",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/yauzl/-/yauzl-2.10.0.tgz",
+      "resolved": "https://registry.npmjs.org/yauzl/-/yauzl-2.10.0.tgz",
       "integrity": "sha1-x+sXyT4RLLEIb6bY5R+wZnt5pfk=",
       "requires": {
         "buffer-crc32": "~0.2.3",


AlexXuChen:
Currently experiencing this issue:
Screenshot from 2021-08-12 08-53-03
How to recreate:
1. Ensure you have an empty or no "xml.fileAssociations" field
2. Open an XML file and open the binding wizard
3. Select the "File association" option, along with any grammar file
4. Reload the window

src/commands/registerCommands.ts
```diff
@@ -181,20 +182,40 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+    const currentFilename = path.basename(identifier.uri);
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
+        await commands.executeCommand(ClientCommandConstants.ASSOCIATE_GRAMMAR_FILE, currentFilename, grammarFilename);
```
angelozerr, line 189: Replace executeCommand with a simple `function addToFileAssociation(currentFilename: string, grammarFilename: string)`. You will be sure that you set string as parameter (and not an Uri for instance). And we don't need to expose `binding with file association` as a command. We need just one entry point `Binding with schema/grammar`

AlexXuChen: In the future, do you think we will want a command just for grammar association, or do you think we will only perform grammar association through the binding wizard and manually editing the settings.json?

src/commands/registerCommands.ts
```diff
@@ -181,20 +182,45 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
+        // Grammar association through setting.json
+        const grammarAssociationObject = {
```
angelozerr, line 190: Here it should be nice to show input to fill a pattern file. This field should be pre filled with **/currentFileName

src/commands/registerCommands.ts
angelozerr: No I dont think. I think its good to have few commands. We will see if we have some user issues about that

src/commands/registerCommands.ts
AlexXuChen: Could you clarify what you mean by "pre filled with **/currentFileName"? Are you referring to adding a regex for the file name?

angelozerr: I dont remember if file association without wildcard with * s working when your xml file is in a folder of workspace root. If it works forget my comment with wildcard but we need an input field pre filled with file name.

Please.do that and it should be really nice to validate this pattern when user is filling this input field to check it matches the file. To do that we need to provide on lemminx side a new command which have the uri (file name) and pattern   (pattern to validate) as parameter and return a boolean to check it the uri matches the given pattern

AlexXuChen: So consider this case, where the xml file is in a different directory from the schema file.

For example, if your directory is as follows:
-dir1
>-file.xsd
>-dir2
>>-file.xml

and you select file.xsd as the associated file for file.xml, the file association object would currently be:
`{
"pattern": "file.xml",
"systemId": file.xsd"
}`

Do you want instead that the user selects the relative path for 'file.xml'? i.e. replaces the ** in **/file.xml?

angelozerr: In this case I would like to have an input field pre-filled with **/file.xml and that's it.

AlexXuChen: If the user doesn't change the **, should the default be the same directory?

src/commands/commandConstants.ts
```diff
@@ -74,6 +74,11 @@ export namespace ClientCommandConstants {
    * Client command to execute an XML command on XML Language Server side.
    */
   export const EXECUTE_WORKSPACE_COMMAND = 'xml.workspace.executeCommand';
+
```
angelozerr, line 77: Remove this unused const

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+    const currentFilename = path.basename(identifier.uri);
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
+        // Grammar association through setting.json
```
angelozerr, line 189: Move this code into a fonction named bindWithFileAssociation

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+          inputPattern = `**/${currentFilename}`;
+        }
+        const grammarAssociationObject = {
+          "pattern": inputPattern,
+          "systemId": grammarFilename
+        }
+        addToValueToSettingArray("xml.fileAssociations", grammarAssociationObject);
+      } else {
+        const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
+        // Insert the proper syntax for binding
```
angelozerr, line 208: Move this code to a function James with bindWithStandard

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+          value: `**/${currentFilename}`,
+          placeHolder: `**/${currentFilename}`,
+          prompt: "Enter the pattern of the XML document(s) to be bound."
+        }
+        let inputPattern = await window.showInputBox(inputBoxOptions);
+        // Default value is **/<currentFilename>
+        if (!inputPattern) {
+          inputPattern = `**/${currentFilename}`;
+        }
+        const grammarAssociationObject = {
```
angelozerr, line 201: Rename the variable with fileAssociation

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
     // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
     const identifier = TextDocumentIdentifier.create(documentURI.toString());
     const grammarURI = fileUri[0];
```
angelozerr, line 184: Move const identifier to the bindWithStandzrd function

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
+        // Grammar association through setting.json
+        const inputBoxOptions = {
+          title: "XML Document Pattern",
+          value: `**/${currentFilename}`,
+          placeHolder: `**/${currentFilename}`,
+          prompt: "Enter the pattern of the XML document(s) to be bound."
+        }
+        let inputPattern = await window.showInputBox(inputBoxOptions);
```
angelozerr, line 196: Replace with 

const inputPattern = (await window.showInputBox(inputBoxOptions)) || `**/${currentFilename}`;

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
+        // Grammar association through setting.json
+        const inputBoxOptions = {
```
angelozerr, line 190: const defaultPattern = `**/${currentFilename}`;

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,58 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+          title: "XML Document Pattern",
+          value: `**/${currentFilename}`,
+          placeHolder: `**/${currentFilename}`,
```
angelozerr, line 193: placeHolder: defaultPattern,

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
+        // Bind grammar using file association
+        await bindWithFileAssociation(grammarFilename, currentFilename);
+      } else {
+        // Bind grammar using standard binding
+        await bindWithStandard(identifier, grammarURI, bindingType, documentURI, languageClient);
```
angelozerr, line 193: Please move identifier declaration inside bindWithStandard

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
 
+/**
+ * Perform grammar binding using file association
+ *
+ * @param grammarFilename the filename of the grammar file
+ * @param currentFilename the filename of the XML document
+ */
+async function bindWithFileAssociation(grammarFilename: string, currentFilename: string) {
+  // Grammar association through setting.json
+  const inputBoxOptions = {
```
angelozerr, line 209: const defaultPattern = `**/${currentFilename}`

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+    title: "XML Document Pattern",
+    value: `**/${currentFilename}`,
```
angelozerr, line 211: value: defaultPattern,

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+    placeHolder: `**/${currentFilename}`,
```
angelozerr, line 212: placeHolder: defaultPattern,

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+    prompt: "Enter the pattern of the XML document(s) to be bound."
+  }
+  let inputPattern = await window.showInputBox(inputBoxOptions);
```
angelozerr, line 215: replace with

```ts
const inputPattern = (await window.showInputBox(inputBoxOptions)) || defaultPattern;
```

src/commands/registerCommands.ts
AlexXuChen: I believe the `currentFileName` requires the identifier. Should I also move it into `bindWithFileAssociation`, or would it be better if I replaced the currentFileName param with the identifier?

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+  // Default value is **/<currentFilename>
+  if (!inputPattern) {
+    inputPattern = `**/${currentFilename}`;
+  }
+  const grammarAssociationObject = {
```
angelozerr, line 220: const pattern = {

src/commands/registerCommands.ts
```diff
@@ -181,22 +182,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+    "pattern": inputPattern,
+    "systemId": grammarFilename
+  }
+  addToValueToSettingArray("xml.fileAssociations", grammarAssociationObject);
```
angelozerr, line 224: addToValueToSettingArray("xml.fileAssociations", fileAssociation);

src/commands/registerCommands.ts
angelozerr: identifier is created from documentURI that bindWithStandard have as parameter. S

replace with 

`await bindWithStandard( grammarURI, bindingType, documentURI, languageClient);`

and inside  bindWithStandard:

`const identifier = TextDocumentIdentifier.create(documentURI.toString());`

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
-    const identifier = TextDocumentIdentifier.create(documentURI.toString());
     const grammarURI = fileUri[0];
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == `fileAssociation`) {
```
angelozerr, line 185: Use single quote

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
     try {
-      const result = await commands.executeCommand(ServerCommandConstants.ASSOCIATE_GRAMMAR_INSERT, identifier, grammarURI.toString(), bindingType);
-      // Insert the proper syntax for binding
-      const lspTextDocumentEdit = <TextDocumentEdit>result;
-      const workEdits = new WorkspaceEdit();
-      for (const edit of lspTextDocumentEdit.edits) {
-        workEdits.replace(documentURI, languageClient.protocol2CodeConverter.asRange(edit.range), edit.newText);
+      if (bindingType == 'fileAssociation') {
+        // Bind grammar using file association
+        await bindWithFileAssociation(grammarURI, documentURI);
```
angelozerr, line 187: Move documentURI before grammarURI because its the document that you bind not the grammar

angelozerr (changes requested):

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+      } else {
+        // Bind grammar using standard binding
+        await bindWithStandard(grammarURI, documentURI, bindingType, languageClient);
```
angelozerr, line 190: Move documentURI before grammarURI

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+ * @param grammarURI the URI of the user selected grammar file
+ * @param documentURI the URI of the current XML document
+ */
+async function bindWithFileAssociation(grammarURI: Uri, documentURI: Uri) {
+  const identifier = TextDocumentIdentifier.create(documentURI.toString());
+  const grammarFilename = path.basename(grammarURI.toString())
+  const currentFilename = path.basename(identifier.uri);
+  const defaultPattern = `**/${currentFilename}`;
+  const inputBoxOptions = {
+    title: "XML Document Pattern",
```
angelozerr, line 210: File Association pattern

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri) {
+  const identifier = TextDocumentIdentifier.create(documentURI.toString());
+  const grammarFilename = path.basename(grammarURI.toString())
+  const currentFilename = path.basename(identifier.uri);
+  const defaultFileAssociationPattern = `**/${currentFilename}`;
```
angelozerr, line 208: Try to keep short name for variable. I would rename to defaultPattern.

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
 }
 
+/**
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri) {
+  const identifier = TextDocumentIdentifier.create(documentURI.toString());
```
angelozerr, line 205: Remove this identifier const.

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+  const grammarFilename = path.basename(grammarURI.toString())
+  const currentFilename = path.basename(identifier.uri);
```
angelozerr, line 207: Replace with:

```ts
const currentFilename = path.basename(documentURI.toString());
```

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
 
+/**
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri) {
+  const identifier = TextDocumentIdentifier.create(documentURI.toString());
+  const grammarFilename = path.basename(grammarURI.toString())
```
angelozerr, line 206: It doesn't work when you select a DTD/XSD which is not in the same folder. The main idea is to store:

 * 1) the URI file path:
 * 
```json
{
      "pattern": "**/bar.xml",
      "systemId": "file:///c%3A/Users/azerr/Downloads/vscode-xml-catalog-test/catalog-preferences/foo.dtd"
    }
```

 * 2) and if the selected XSD/DTD comes from the workspace we should store the absolute path from workspace root.
```json
{
      "pattern": "**/bar.xml",
      "systemId": "/catalog-preferences/foo.dtd"
    }
```

To support 1), replace with 

`const grammarFilePath = grammarURI.toString();`

src/commands/registerCommands.ts
```diff
@@ -179,24 +180,81 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+  const inputBoxOptions = {
+    title: "File Association Pattern",
+    value: defaultPattern,
+    placeHolder: defaultPattern,
+    prompt: "Enter the pattern of the XML document(s) to be bound."
+  }
+  const inputPattern = (await window.showInputBox(inputBoxOptions)) || defaultPattern;
+  const fileAssociation = {
+    "pattern": inputPattern,
+    "systemId": grammarFilename
```
angelozerr, line 218: "systemId": grammarFilePath

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,83 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri) {
+  const grammarFilePath = grammarURI.toString();
+  const currentFilename = path.basename(documentURI.toString());
+  const currentWorkspaceUri = getAbsoluteCurrentWorkspaceUri();
+  const relativeGrammarFilePath = getRelativeUri(currentWorkspaceUri, grammarFilePath);
```
angelozerr, line 209: If user select a xsd which is not inside workspace it should return the uri whith file:// is it working like this?


I will rename relatibegrammarfilepath with grammarFilePath singe the path can be relative to workspace or not

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,50 @@
+import * as path from "path";
+import { TextDocument, window, workspace, WorkspaceFolder } from "vscode";
+
+/**
+ * Return the current workspace uri from root
+ *
+ * @returns the path from root of the current workspace
+ */
+export function getCurrentWorkspaceUri(): string {
```
angelozerr, line 9: I will rename the function with getWorkspaceUri(fileUri: Uri)

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Return the relative file path between a start and destination uri
+ *
+ * @param startPath the absolute path of the beginning directory
+ * @param destinationPath the absolute path of destination file
+ * @returns the path to the destination relative to the start
+ */
+export function getRelativeUri(startPath: string, destinationPath: string): string {
+  const relativePath = path.relative(path.normalize(startPath), path.normalize(destinationPath));
+  return relativePath
```
angelozerr, line 49: return path.relative....

src/commands/registerCommands.ts
AlexXuChen: From https://github.com/redhat-developer/vscode-xml/pull/584#discussion_r690194431, at the moment the implemented behaviour is ii (the relative path). Did you want i (the absolute path) instead?

AlexXuChen: in your example for ii, folder catalog-preferences inside the workspace?

AlexXuChen: And would it be sufficient to just use the relative path, or would you like the i and ii cases?

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Return the uri of the current file
+ *
+ * @returns the uri of the current file
+ */
+export function getCurrentFileUri(): string {
+  const currentFile: TextDocument = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ?  window.activeTextEditor.document : undefined;
+  const currentFileUri: string = (currentFile && currentFile.uri) ? currentFile.uri.fsPath : undefined;
```
angelozerr, line 36: Return directly without using  a const variable

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+    || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath);
+  return currentWorkspaceUri;
+}
+
+/**
+ * Return the current workspace absolute uri
+ *
+ * @returns the absolute path of the current workspace
+ */
+export function getAbsoluteCurrentWorkspaceUri(): string {
```
angelozerr, line 22: Rename to getAbsoluteWorkspaceUri(fileUri: uri)

src/commands/registerCommands.ts
AlexXuChen: I just made a fix where the grammarFilePath now depends on if the file is in/out of the workspace folder

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+import { TextDocument, window, workspace, WorkspaceFolder } from "vscode";
+
+/**
+ * Return the current workspace uri from root
+ *
+ * @returns the path from root of the current workspace
+ */
+export function getWorkspaceUri(): string {
+  const currentFile: TextDocument = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ?  window.activeTextEditor.document : undefined;
+  const currentWorkspace: WorkspaceFolder = (currentFile && currentFile.uri) ?  workspace.getWorkspaceFolder(currentFile.uri) : undefined;
```
angelozerr, line 11: Please move this currentFile in the function parameter and do like this for other function.

angelozerr (changes requested):

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,84 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
 
+/**
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri) {
+  const absoluteGrammarFilePath = grammarURI.toString();
+  const currentFilename = path.basename(documentURI.toString());
```
angelozerr, line 207: Dont use path here but à function of fileutils

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+  return ((currentWorkspace && currentWorkspace.uri) || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri)).toString();
+}
+
+/**
+ * Return the uri of the current file
+ *
+ * @returns the uri of the current file
+ */
+export function getFileUri(): string {
```
angelozerr, line 31: Rename with getCurrentFileUri

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+  const currentWorkspace: WorkspaceFolder = (currentFile && currentFile.uri) ?  workspace.getWorkspaceFolder(currentFile.uri) : undefined;
+ return ((currentWorkspace && currentWorkspace.uri.fsPath) || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath));
+}
+
+/**
+ * Return the current workspace absolute uri
+ *
+ * @returns the absolute path of the current workspace
+ */
+export function getAbsoluteWorkspaceUri(): string {
```
angelozerr, line 20: Move currentFile in the function parameter

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,84 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+  const currentWorkspaceUri = getAbsoluteWorkspaceUri();
+  const relativeGrammarFilePath = getRelativeUri(currentWorkspaceUri, absoluteGrammarFilePath);
+  const grammarFilePath = absoluteGrammarFilePath.includes(currentWorkspaceUri) ? relativeGrammarFilePath : absoluteGrammarFilePath;
```
angelozerr, line 210: I think currentWorkspaceUri can be undefined are you sur that we can call includes with undefined?

AlexXuChen: if currentWorkspaceUri is undefined, it would make grammarFilePath the absoluteFilePath in line 211 correct?

AlexXuChen: Ah i understand your question now you are referring to absoluteGrammarFilePath.includes(currentWorkspaceUri). If currentWorkspaceUri is undefined, it will be false, and the grammar file will use the absolute path. To me this makes sense since this can happen when a workspace is not open (i.e. a single open editor), so no relative path can be formed and therefore the absolute path should always be used.

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,86 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ * @param currentFile the opened TextDocument
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, currentFile: TextDocument) {
+  const absoluteGrammarFilePath = grammarURI.toString();
+  const currentFilename = getCurrentFileName(documentURI.toString());
+  const currentWorkspaceUri = getAbsoluteWorkspaceUri(currentFile);
+  const relativeGrammarFilePath = getRelativeUri(currentWorkspaceUri, absoluteGrammarFilePath);
+  const grammarFilePath = absoluteGrammarFilePath.includes(currentWorkspaceUri) ? relativeGrammarFilePath : absoluteGrammarFilePath;
```
AlexXuChen, line 212: If I look at a directory path and a file path, where the file has the same name as the directory, but is outside the directory, e.g:
a/b/c/foo -> directory
a/b/c/foo.xml -> file, outside of above directory
In this case, the includes will be true and it will use the relative path when it shouldn't.
The solution here is to only compare the directories instead of the uri with the file

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,86 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+/**
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ * @param currentFile the opened TextDocument
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, currentFile: TextDocument) {
+  const absoluteGrammarFilePath = grammarURI.toString();
+  const currentFilename = getCurrentFileName(documentURI.toString());
```
angelozerr, line 209: Rename with getFileName

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+  return (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ?  window.activeTextEditor.document : undefined;
+}
+
+/**
+ * Return the uri of the current file
+ *
+ * @param currentFile the opened TextDocument
+ * @returns the uri of the current file
+ */
+export function getCurrentFileUri(currentFile: TextDocument): string {
```
angelozerr, line 41: Rename with getFileUri

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+  return (currentFile && currentFile.uri) ? currentFile.uri.fsPath : undefined;
+}
+
+/**
+ * Uses path to return a basename from a uri
+ *
+ * @param filePath the uri of the file
+ * @return the filename
+ */
+export function getCurrentFileName(filePath: string): string {
```
angelozerr, line 51: Rename with getFileName

src/utils/fileUtils.ts
AlexXuChen: This contradicts https://github.com/redhat-developer/vscode-xml/pull/584#discussion_r690658037, is that alright?

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,76 @@
+import * as path from "path";
+import { TextDocument, window, workspace, WorkspaceFolder } from "vscode";
+
+/**
+ * Return the current workspace uri from root
+ *
+ * @param currentFile the opened TextDocument
+ * @returns the path from root of the current workspace
+ */
+export function getWorkspaceUri(currentFile: TextDocument): string {
```
angelozerr, line 10: Rename currentFile with document and do that for other function which have textdocument

src/utils/fileUtils.ts
AlexXuChen: @angelozerr

angelozerr: Yes sorry rename with getFilePath

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Return the relative file path between a start and destination uri
+ *
+ * @param startPath the absolute path of the beginning directory
+ * @param destinationPath the absolute path of destination file
+ * @returns the path to the destination relative to the start
+ */
+export function getRelativeUri(startPath: string, destinationPath: string): string {
+  const relative = path.relative(path.normalize(startPath), path.normalize(destinationPath));
```
angelozerr, line 63: Rename with getRelativePath ans rename directly the value

AlexXuChen: Sorry, forgot to remove after debug

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+  return (document && document.uri) ? document.uri.fsPath : undefined;
+}
+
+/**
+ * Uses path to return a basename from a uri
+ *
+ * @param filePath the uri of the file
+ * @return the filename
+ */
+export function getCurrentFileName(filePath: string): string {
```
angelozerr, line 51: Rename with getFileName

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+  const currentWorkspace: WorkspaceFolder = (document && document.uri) ? workspace.getWorkspaceFolder(document.uri) : undefined;
+  return ((currentWorkspace && currentWorkspace.uri) || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri)).toString();
+}
+
+/**
+ * Retrieve the text document that is open in the editor
+ *
+ * @returns the text document open in the editor
+ */
+export function getCurrentFile(): TextDocument {
```
angelozerr, line 31: Remove this function

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+  return (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ? window.activeTextEditor.document : undefined;
+}
+
+/**
+ * Return the uri of the current file
+ *
+ * @param document the opened TextDocument
+ * @returns the uri of the current file
+ */
+export function getCurrentFileUri(document: TextDocument): string {
```
angelozerr, line 41: Rename -ith getCurrentFilePath() without param

src/utils/fileUtils.ts
AlexXuChen: This function takes out the vscode import(s) in variableSubstitution.ts, should I still remove it?

src/utils/fileUtils.ts
AlexXuChen: By removing the param, this function loses value. Should I just put the logic back into variable substitution?

src/utils/fileUtils.ts
AlexXuChen: This function removes the vscode imports in variableSubstitution.ts. Should I still remove it?

angelozerr:
I have not seen the code but when xsd is in the workspace it generates (on Windows OS) path with  \\ instead of /


 ```json
{
  "pattern": "**/foo2.xml",
  "systemId": "catalog-preferences\\foo.dtd"
}
```

It should be nice to have 

```json
{
  "pattern": "**/foo2.xml",
  "systemId": "catalog-preferences/foo.dtd"
}
```

AlexXuChen:
> ```json
> {
>  "pattern": "**/foo2.xml",
>  "systemId": "catalog-preferences\\foo.dtd"
> }
> ```

Does this only happen with grammar files in the same workspace?

AlexXuChen:
I just made a change to make the relative path use POSIX. If the absolute path also uses \\ in Windows, I can make the change for absolute as well.

angelozerr:
I tried to bind a foo2.xml with a foo2.dtd which is in the same folder and the system id have a bad file path:

```
{
      "pattern": "**/foo2.xml",
      "systemId": "../file:\\c%3A\\Users\\azerr\\Downloads\\vscode-xml-catalog-test\\catalog-preferences\\foo2.dtd"
    }
```

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,86 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+  const grammarFilePath = getDirectoryPath(absoluteGrammarFilePath).includes(currentWorkspaceUri) ? getRelativePath(currentWorkspaceUri, absoluteGrammarFilePath) : absoluteGrammarFilePath;
+
+  const defaultPattern = `**/${currentFilename}`;
+  const inputBoxOptions = {
+    title: "File Association Pattern",
+    value: defaultPattern,
+    placeHolder: defaultPattern,
+    prompt: "Enter the pattern of the XML document(s) to be bound."
+  }
+  const inputPattern = (await window.showInputBox(inputBoxOptions)) || defaultPattern;
```
angelozerr, line 221: Replace with:

```ts
const inputPattern = await window.showInputBox(inputBoxOptions);
  if (!inputPattern) {
    // User closed the input box with Esc
    return;
  }
```

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * Return the relative file path between a start and destination uri
+ *
+ * @param startPath the absolute path of the beginning directory
+ * @param destinationPath the absolute path of destination file
+ * @returns the path to the destination relative to the start
+ */
+export function getRelativePath(startPath: string, destinationPath: string): string {
```
angelozerr, line 53: Replace with:

```ts
export function getRelativePath(startPath: string, destinationPath: string): string {
  return path.relative(path.normalize(startPath), path.normalize(destinationPath)).replace(/\\/g, '/');
}
```

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+  return ((currentWorkspace && currentWorkspace.uri.fsPath) || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath));
+}
+
+/**
+ * Return the current workspace absolute uri
+ *
+ * @param document the opened TextDocument
+ * @returns the absolute path of the current workspace
+ */
+export function getAbsoluteWorkspaceUri(document: TextDocument): string {
```
angelozerr, line 21: Remove this function

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,66 @@
+import * as path from "path";
+import { TextDocument, window, workspace, WorkspaceFolder } from "vscode";
+
+/**
+ * Return the current workspace uri from root
+ *
+ * @param document the opened TextDocument
+ * @returns the path from root of the current workspace
+ */
+export function getWorkspaceUri(document: TextDocument): string {
```
angelozerr, line 10: Replace with :

```ts
export function getWorkspaceUri(document: TextDocument): Uri | undefined {
  const currentWorkspace: WorkspaceFolder = (document && document.uri) ? workspace.getWorkspaceFolder(document.uri) : undefined;
  return ((currentWorkspace && currentWorkspace.uri) || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri));
}
```

src/commands/registerCommands.ts
```diff
@@ -179,24 +181,86 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
...
+ * Perform grammar binding using file association through settings.json
+ *
+ * @param documentURI the URI of the current XML document
+ * @param grammarURI the URI of the user selected grammar file
+ * @param document the opened TextDocument
+ */
+async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, document: TextDocument) {
+  const absoluteGrammarFilePath = grammarURI.toString();
+  const currentFilename = getFileName(documentURI.toString());
+  const currentWorkspaceUri = getAbsoluteWorkspaceUri(document);
```
angelozerr, line 210: Replace with:

```ts
const currentWorkspaceUri = getWorkspaceUri(document);
  if (!currentWorkspaceUri) {
    return;
  }
  // If the grammar file is in the same workspace, use the relative path, otherwise use the absolute path
  const grammarFilePath = getDirectoryPath(absoluteGrammarFilePath).includes(currentWorkspaceUri.toString()) ? getRelativePath(currentWorkspaceUri.toString(), absoluteGrammarFilePath) : absoluteGrammarFilePath;

```

src/settings/variableSubstitution.ts
```diff
@@ -99,11 +100,9 @@ const VARIABLE_SUBSTITUTIONS: VariableSubstitution[] = [
...
 
   // Collect properties needed to resolve variables
-  const currentFile: TextDocument = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ?  window.activeTextEditor.document : undefined;
-  const currentFileUri: string = (currentFile && currentFile.uri) ? currentFile.uri.fsPath : undefined;
-  const currentWorkspace: WorkspaceFolder = (currentFile && currentFile.uri) ?  workspace.getWorkspaceFolder(currentFile.uri) : undefined;
-  const currentWorkspaceUri: string = (currentWorkspace && currentWorkspace.uri.fsPath)
-    || (workspace.workspaceFolders && workspace.workspaceFolders[0].uri.fsPath);
+  const currentFile = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ? window.activeTextEditor.document : undefined;
+  const currentFileUri: string = getFilePath(currentFile);
+  const currentWorkspaceUri: string = getWorkspaceUri(currentFile);
```
angelozerr, line 105: Replace with

`const currentWorkspaceUri= getWorkspaceUri(currentFile);`

angelozerr: And line 25, replace with

`      newVal = settingVariable.substituteString(newVal, currentFileUri, currentWorkspaceUri && currentWorkspaceUri.toString());`

angelozerr:
@AlexXuChen could you squash your 2 commits in one commit please.

angelozerr (approved):

angelozerr:
It works like a charm. Great job @AlexXuChen !

PR #592, opened 2021-09-03
Validate the file assocation pattern in the wizard

### LINKED ISSUE #586: File association pattern validation
Related to #573.

When a grammar file is associated with an XML document using `xml.fileAssociations`, we should ensure that the pattern of the XML document file(s) (`pattern` field) is valid.

For example, when the file association pattern input is accepted:
Screenshot from 2021-08-18 13-55-45
And the text entry is an invalid pattern, no file association similar to the example below should be entered:
```
{
"pattern": "**/<invalid pattern>.xml",
"systemId": "<some grammar file path>
}
```
and instead there should be an error thrown with no file association.

rgrunber:
- Fixes #586
- Declare command "xml.check.file.pattern" to be used on the server side
  to determine if a given pattern will match any files

This will require https://github.com/eclipse/lemminx/pull/1112 .

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -216,7 +216,16 @@ async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, docume
     title: "File Association Pattern",
     value: defaultPattern,
     placeHolder: defaultPattern,
-    prompt: "Enter the pattern of the XML document(s) to be bound."
+    prompt: "Enter the pattern of the XML document(s) to be bound.",
+    validateInput: async (pattern: string) => {
+      let hasMatch = false;
+      try {
+        hasMatch = await commands.executeCommand(ClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, ServerCommandConstants.CHECK_FILE_PATTERN, pattern, documentURI.toString());
+      } catch (error) {
+        console.log(`Error while validating file pattern : ${error}`);
+      }
+      return !hasMatch ? "The pattern will not match any file." : null
+    }
   }
   const inputPattern = (await window.showInputBox(inputBoxOptions));
   if (!inputPattern) {
--- a/src/commands/commandConstants.ts
+++ b/src/commands/commandConstants.ts
@@ -102,4 +102,9 @@ export namespace ServerCommandConstants {
    * Command to check if the current XML document is bound to a grammar
    */
   export const CHECK_BOUND_GRAMMAR = "xml.check.bound.grammar"
+
+  /**
+   * Command to check if a given file pattern matches any file on the workspace
+   */
+  export const CHECK_FILE_PATTERN = "xml.check.file.pattern"
 }


src/commands/registerCommands.ts
```diff
@@ -216,7 +216,16 @@ async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, docume
     title: "File Association Pattern",
     value: defaultPattern,
     placeHolder: defaultPattern,
-    prompt: "Enter the pattern of the XML document(s) to be bound."
+    prompt: "Enter the pattern of the XML document(s) to be bound.",
+    validateInput: async (input: string) => {
+      var hasMatch = false;
+      try {
+        hasMatch = await commands.executeCommand(ClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, ServerCommandConstants.CHECK_FILE_PATTERN, input);
```
angelozerr, line 223: You should set the uri as args of the current xml file

src/commands/registerCommands.ts
```diff
@@ -216,7 +216,16 @@ async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, docume
     title: "File Association Pattern",
     value: defaultPattern,
     placeHolder: defaultPattern,
-    prompt: "Enter the pattern of the XML document(s) to be bound."
+    prompt: "Enter the pattern of the XML document(s) to be bound.",
+    validateInput: async (input: string) => {
+      var hasMatch = false;
```
angelozerr, line 221:
```ts
let hasMatch = false;
```

src/commands/registerCommands.ts
```diff
@@ -216,7 +216,16 @@ async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, docume
     title: "File Association Pattern",
     value: defaultPattern,
     placeHolder: defaultPattern,
-    prompt: "Enter the pattern of the XML document(s) to be bound."
+    prompt: "Enter the pattern of the XML document(s) to be bound.",
+    validateInput: async (input: string) => {
```
angelozerr, line 220: rename `input `with `pattern`

src/commands/registerCommands.ts
```diff
@@ -216,7 +216,16 @@ async function bindWithFileAssociation(documentURI: Uri, grammarURI: Uri, docume
...
+      var hasMatch = false;
+      try {
+        hasMatch = await commands.executeCommand(ClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, ServerCommandConstants.CHECK_FILE_PATTERN, input, documentURI.toString());
+      } catch (error) {
+        console.log(`Error while validating file pattern : ${error}`);
```
angelozerr, line 225: If I understand, when there is an error on language server side, it means that validate return false and wizard cannot be used. It it the expected behavior?

rgrunber: Yes, that's how it should work. Any error in the language server side, or even a failure in the actual executeCommand. I can quickly try to confirm.

rgrunber: Yup, `message:'Running the contributed command: 'xml.workspace.executeCommand' failed.'` if the command on the language server side throws some kind of exception or fails in some unexpected way.

angelozerr (approved):

angelozerr:
It works like a charm, many thanks @rgrunber !

PR #606, opened 2021-10-18
Prevent from workspace uri undefined

angelozerr:
Prevent from workspace uri undefined


rgrunber:
@AlexXuChen was able to reproduce this (myself as well) when only a file (no workspace) is opened, and attempting to associate that xml file with a schema on the filesystem.

PR #640, opened 2022-01-26
Download external resources preference

angelozerr:
Download external resources preference

### DIFF
--- a/src/settings/settings.ts
+++ b/src/settings/settings.ts
@@ -226,8 +226,10 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
   const isWorkspaceTrusted = (workspace as any).isTrusted;
   if (isWorkspaceTrusted !== undefined && !isWorkspaceTrusted) {
     xml['xml']['validation']['resolveExternalEntities'] = false;
+    xml['xml']['downloadExternalResources']['enabled'] = false;
   } else {
     xml['xml']['validation']['resolveExternalEntities'] = workspace.getConfiguration('xml').get('validation.resolveExternalEntities', false);
+    xml['xml']['downloadExternalResources']['enabled'] = workspace.getConfiguration('xml').get('downloadExternalResources.enabled', false);
   }
 
   //applying externalXmlSettings to the xmlSettings
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ The following settings are supported:
 * `xml.trace.server`: Trace the communication between VS Code and the XML language server in the Output view. Default is `off`.
 * `xml.logs.client`: Enable/disable logging to the Output view. Default is `true`.
 * [`xml.catalogs`](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Preferences.md#catalogs): Register XML catalog files.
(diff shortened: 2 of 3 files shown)


src/settings/settings.ts
```diff
@@ -226,8 +226,10 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
   const isWorkspaceTrusted = (workspace as any).isTrusted;
   if (isWorkspaceTrusted !== undefined && !isWorkspaceTrusted) {
```
rgrunber, line 227: If you insert a breakpoint here, you'll see there are 2 instances of this settings :

`xml.preferences.downloadExternalRessource`
`xml.downloadExternalResource`

Just make sure whichever namespace is chosen, it is the one the language server is able to parse.

angelozerr: I change and fix lemminx with `xml.downloadExternalResources`

src/settings/settings.ts
```diff
@@ -225,9 +225,11 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
   // Check workspace trust
   const isWorkspaceTrusted = (workspace as any).isTrusted;
   if (isWorkspaceTrusted !== undefined && !isWorkspaceTrusted) {
-    xml['xml']['validation']['resolveExternalEntities'] = false;
+    xml['xml']['validation']['resolveExternalEntities']['enabled'] = false;
+    xml['xml']['downloadExternalResources'] = false;
```
fbricon, line 229: xml['xml']['downloadExternalResources']['enabled'] = false;

angelozerr: fixed

src/settings/settings.ts
```diff
@@ -225,9 +225,11 @@ export function getXMLSettings(javaHome: string | undefined, logfile: string, ex
   // Check workspace trust
   const isWorkspaceTrusted = (workspace as any).isTrusted;
   if (isWorkspaceTrusted !== undefined && !isWorkspaceTrusted) {
-    xml['xml']['validation']['resolveExternalEntities'] = false;
+    xml['xml']['validation']['resolveExternalEntities']['enabled'] = false;
```
rgrunber, line 228: You would need to also update the settings in `package.json` and in the LemMinX settings structure.

Currently I'm seeing : `Activating extension 'redhat.vscode-xml' failed: Cannot create property 'enabled' on boolean 'false'.`

angelozerr: Sorry for this mistake. The correct settings is `xml.resolveExternalEntities`. I fixedit.

rgrunber:
One small case I noticed :

1) Open a workspace and do not grant trust. Remote schemas won't download, as expected.
2) Now grant workspace trust
3) The schema still won't download, and will show a message for the download being disabled
4) If you reload the window it should work as expected again

This is because workspace trust is checked once at activation, but we're not using the `onDidGrantWorkspaceTrust` event. See https://github.com/redhat-developer/vscode-java/blob/master/src/extension.ts#L447-L449 for how to do it. I guess this may also be an issue with resolving entities ?

rgrunber (approved):
I'm fine with merging. @AlexXuChen , can address the issue I mentioned in a separate PR.

fbricon:
thanks @angelozerr!

PR #642, opened 2022-01-28
Added `onDidGrantWorkspaceTrust` event for `downloadExternalResource` setting

AlexXuChen:
Added `onDidGrantWorkspaceTrust` event for `downloadExternalResource` setting.

Fixes https://github.com/redhat-developer/vscode-xml/pull/640#issuecomment-1023701483

Requires #640 

### DIFF
--- a/src/client/xmlClient.ts
+++ b/src/client/xmlClient.ts
@@ -103,6 +103,14 @@ export async function startLanguageClient(context: ExtensionContext, executable:
     }
   }));
 
+  const onDidGrantWorkspaceTrust = (workspace as any).onDidGrantWorkspaceTrust;
+  if (onDidGrantWorkspaceTrust !== undefined) {
+    context.subscriptions.push(onDidGrantWorkspaceTrust(() => {
+      languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getXMLSettings(requirementsData.java_home, logfile, externalXmlSettings) });
+      workspace.getConfiguration('xml').update('downloadExternalResources.enabled', true); //set back to default setting
+    }));
+  }
+
   return languageClient;
 }
 
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
         "xml.extension.jars",
         "xml.java.home",
         "xml.server.vmargs",
-        "xml.validation.resolveExternalEntities"
+        "xml.validation.resolveExternalEntities",
+        "xml.downloadExternalResources.enabled"
       ]
     }
   },


src/extension.ts
```diff
@@ -64,6 +64,13 @@ export async function activate(context: ExtensionContext): Promise<XMLExtensionA
 
   languageClient = await startLanguageClient(context, serverOptions, logfile, externalXmlSettings, requirementsData);
 
+  const onDidGrantWorkspaceTrust = (workspace as any).onDidGrantWorkspaceTrust;
```
angelozerr, line 67: The `onDidGrantWorkspaceTrust` should be managed in xmlClient at https://github.com/redhat-developer/vscode-xml/blob/ed8e6bb8c4bc686144870d58f7dfb7f9dd4715e5/src/client/xmlClient.ts#L105

src/extension.ts
```diff
@@ -64,6 +64,13 @@ export async function activate(context: ExtensionContext): Promise<XMLExtensionA
...
+  if (onDidGrantWorkspaceTrust !== undefined) {
+    context.subscriptions.push(onDidGrantWorkspaceTrust(() => {
+      commands.executeCommand("xml.downloadExternalResources.enabled", true);
```
angelozerr, line 70: I think the proper code should be the same thanblock code at https://github.com/redhat-developer/vscode-xml/blob/ed8e6bb8c4bc686144870d58f7dfb7f9dd4715e5/src/client/xmlClient.ts#L150

I suggest that you move this blick in a private function inXMLClient.ts because I noticed that this code is used in:

 * https://github.com/redhat-developer/vscode-xml/blob/master/src/client/xmlClient.ts#L150
 * https://github.com/redhat-developer/vscode-xml/blob/ed8e6bb8c4bc686144870d58f7dfb7f9dd4715e5/src/client/xmlClient.ts#L101

and now we your onDidGrantWorkspaceTrust too.

rgrunber: `onDidGrantWorkspaceTrust` as its name implies is only triggered when trust is granted (so not when something changes to untrusted). `(workspace as any).isTrusted` is `true` in the callback though.

angelozerr: OK so it means my suggestion was wrong and we need to update settings as Alex did the first time?

rgrunber: Was just going to say. the setting, `xml.downloadExternalResources.enabled` as `true` needs to be injected. The problem with the previous path was that it isn't a command to be executed.

AlexXuChen: Can we inject the setting change here:
https://github.com/redhat-developer/vscode-xml/blob/ed8e6bb8c4bc686144870d58f7dfb7f9dd4715e5/src/settings/settings.ts#L227-L232
since we generate new XML settings from the `onDidGrantWorkspaceTrust` event?

rgrunber: If we set it to true based on the the workspace trust state, then we can never have a case where a user sets it to false, which might be a disadvantage.

Let's just manually set it in the callback.

src/client/xmlClient.ts
```diff
@@ -103,6 +103,14 @@ export async function startLanguageClient(context: ExtensionContext, executable:
     }
   }));
 
+  const onDidGrantWorkspaceTrust = (workspace as any).onDidGrantWorkspaceTrust;
+  if (onDidGrantWorkspaceTrust !== undefined) {
+    context.subscriptions.push(onDidGrantWorkspaceTrust(() => {
+      languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getXMLSettings(requirementsData.java_home, logfile, externalXmlSettings) });
+      workspace.getConfiguration('xml').update('downloadExternalResources.enabled', true); //set back to default setting
```
AlexXuChen, line 110: I set it back to the initial default, but I suppose I could also store the setting in `externalXMLSettings` and retrieve once workspaceTrust is renabled?

AlexXuChen:
https://github.com/redhat-developer/vscode-xml/pull/640#issuecomment-1023701483 should behave as expected now:
workspacetrustdownloadexternal

src/client/xmlClient.ts
rgrunber: `externalXmlSettings` are settings other extensions might contribute, so I don't think it's a good place. I think it's fine as-is.

src/extension.ts
```diff
@@ -13,7 +13,7 @@
 import * as fs from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { ExtensionContext, extensions, languages } from "vscode";
+import { commands, ExtensionContext, extensions, languages, workspace } from "vscode";
```
rgrunber, line 16: These 2 additions can be removed, as they seem unused now.

rgrunber (approved):
LGTM.

PR #644, opened 2022-02-01
Use LemMinX 0.18.3

rgrunber:

CC @angelozerr . We need this just prior to building to use the released LS.


angelozerr (approved):

angelozerr:
Thanks @rgrunber

PR #650, opened 2022-02-03
Eliminate need for downloading just-released binary checksums.

rgrunber:
- When the build process finishes building binaries, and generating the
  checksums, we should stash these to avoid having to download once
  deployed
- Also use ovsx<0.3.0 to ensure we build with Node v12


AlexXuChen (approved):
LGTM

PR #660, opened 2022-02-14
Upversion to 0.19.1

rgrunber:

@angelozerr In the future, don't forget that the `package-lock.json` also needs to be updated. This didn't affect the release as `0.19.0` was set everywhere, but good to update it for consistency.


AlexXuChen (approved):
LGTM

PR #685, opened 2022-03-28
Add support to download binaries from GitHub Releases.

rgrunber:
- Migrate LemMinX binary server from download.jboss.org to github.com
- GitHub Release URL returns 302 and does not properly set Content-Type
  so we should handle these cases
- Archive the zipped binaries, checksums and vsix as a temporary measure
  until GH Release API can be used


angelozerr (approved):
LGTM

PR #695, opened 2022-04-05
Added permission check and window error message for LemMinx binary

AlexXuChen:
Added permission check and window error message for LemMinx binary

Peek 2022-04-07 10-00

### DIFF
--- a/src/server/binary/binaryServerStarter.ts
+++ b/src/server/binary/binaryServerStarter.ts
@@ -64,9 +64,16 @@ async function getServerBinaryPath(): Promise<string> {
   const binaryPath: string = config.get("server.binary.path");
   if (binaryPath) {
     if (fs.existsSync(binaryPath)) {
-      return Promise.resolve(binaryPath);
+      try {
+        // checks read/execute permissions
+        fs.accessSync(binaryPath, fs.constants.R_OK | fs.constants.X_OK)
+        return Promise.resolve(binaryPath);
+      } catch(e) {
+        window.showErrorMessage('Permission(s) denied for the specified XML language server binary. Using the default binary...');
+      }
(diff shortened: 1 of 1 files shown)


rgrunber:
There's actually nothing we can do to fix this on the build side for binaries uploaded using the upload/download actions : https://github.com/actions/upload-artifact#permission-loss .

src/server/binary/binaryServerStarter.ts
```diff
@@ -64,9 +64,19 @@ async function getServerBinaryPath(): Promise<string> {
   const binaryPath: string = config.get("server.binary.path");
   if (binaryPath) {
     if (fs.existsSync(binaryPath)) {
-      return Promise.resolve(binaryPath);
+      try {
+        // checks execute permission
+        fs.accessSync(binaryPath, fs.constants.X_OK)
+        // checks read/write permission
+        fs.accessSync(binaryPath, fs.constants.R_OK | fs.constants.W_OK)
```
rgrunber, line 71: Can the 2 lines be combined into one ? Also we don't need write permissions for this to work so I would just remove it.

```suggestion
        // checks read/execute permission
        fs.accessSync(binaryPath, fs.constants.R_OK | fs.constants.X_OK)
```

rgrunber: I would also remove the `console.error(..)`.

AlexXuChen: Do you think there's any value in have 2 separate messages for permissions (i.e. one for executable permissions and one for read? Otherwise I can combine both, and leave the generic permission denied message.

rgrunber (approved):
Change looks good to me.

PR #706, opened 2022-04-29
Add LemMinX binary artifacts to release tab only on new changes.

rgrunber:
- Save the last commit from LemMinX in a file (lastCommit)
- Use the GH cache action to cache the file
- If a cache hit occurs, then the commit has been built, otherwise it is
  a new commit and should result in releasing the binary artifacts


angelozerr (approved):
It looks good.

PR #707, opened 2022-05-07
Replace obsolete AdoptOpenJDK link with Adoptium

claui:
In 2021, [AdoptOpenJDK has become Adoptium](https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/).

The old site does point to the new site but that requires several additional clicks to get there. Besides, the [old download site](https://adoptopenjdk.net/releases.html) is no longer being updated, and fails to mention JDK 17.

See also #270.


rgrunber (approved):
Good catch. The fact that latest is still OpenJDK 16 there is a hint we should have changed.

Same could probably be done for JDK 11 link at https://github.com/redhat-developer/vscode-microprofile/blob/master/CONTRIBUTING.md#installation-prerequisites .

PR #712, opened 2022-05-19
Move to vscode-languageclient 8.0.1

### LINKED ISSUE #711: Move to vscode-languageclient 8.0.1
With vscode-languageclient 8.0.1, we will benefit with:

 * LSP 3.7 support
 * we could remove shouldLanguageServerExitOnShutdown

angelozerr:
Move to vscode-languageclient 8.0.1

Fixes #711

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -172,7 +172,7 @@
     },
     "@types/yauzl": {
       "version": "2.9.1",
-      "resolved": "https://registry.npmjs.org/@types/yauzl/-/yauzl-2.9.1.tgz",
+      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/@types/yauzl/-/yauzl-2.9.1.tgz",
       "integrity": "sha512-A1b8SU4D10uoPjwb0lnHmmu8wZhR9d+9o2PKBQT2jU5YPTKsxac6M2qGAdY7VcL+dHHhARVUDmeg0rOrcd9EjA==",
       "dev": true,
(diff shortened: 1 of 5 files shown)


angelozerr:
With this PR we need to test that:

 * XML features should work (completion, validation, hover, etc).
 * when we close the vscode, we must not have some zombie process (because I remove shouldLanguageServerExitOnShutdown)
 * when we change settings, we must test that it continues to work (refresh of settings should continue to work).

AlexXuChen (approved):
LGTM, XML features and settings are still working, and language server shuts down on close:
Peek 2022-05-19 10-02

rgrunber:
Don't forget about the server-side property : https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java#L230-L232

angelozerr:
> Actually, I guess we can keep it for other clients ? It doesn't hurt to keep.

Indeed, it is rhe reason why I keep it.

> I can also confirm the behaviour is of shouldLanguageServerExitOnShutdown is enforced automatically in 8.0.1 without using the property. Note that we always would shut down the language server eventually (30s after close thanks to the parent process watcher, in the worst case), but this change preserves the immediate shutdown behaviour.

If I understand correctly, it works for you?

rgrunber:
Yes, works for me :)

angelozerr:
Nice, thanks for the fedback!

PR #713, opened 2022-05-19
TextMate grammar for *.dtd files

### LINKED ISSUE #313: TextMate grammar for *.dtd files
I think it could be helpful to provide a TextMate grammar for `*.dtd` files. Syntax highlighting would make it a bit easier to read and edit these files.

It looks like nothing is being done to implement it upstream in VSCode: [VSCode related issue](https://github.com/microsoft/vscode/issues/55751). VSCode just reuses the Atom TextMate grammar for XML/DTD: [Atom issue](https://github.com/atom/language-xml/issues/89). The Atom TextMate grammar was extracted from the [TextMate XML plugin](https://github.com/textmate/xml.tmbundle). This grammar doesn't include syntax highlighting for DTD.

angelozerr:
TextMate grammar for *.dtd files

Fixes #313

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "activationEvents": [
     "onLanguage:xml",
     "onLanguage:xsl",
+    "onLanguage:dtd",
     "onCommand:xml.open.docs",
     "onCommand:xml.open.docs.home"
   ],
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -97,6 +97,27 @@
     "yauzl": "^2.10.0"
   },
   "contributes": {
+    "languages": [
+      {
+        "id": "dtd",
```
AlexXuChen, line 102: We also need to add the "onLanguage:dtd" to "activationEvents" since if we open a `.dtd` file first, XML Support won't start.

AlexXuChen (changes requested):

package.json
angelozerr: Good catch! Fixed.

AlexXuChen (approved):
LGTM

angelozerr:
I merged the PR it is a good start, but it requires a lot of improvement. We will improve it step by step and if we have good result, we could try to contribute to vscode.

PR #715, opened 2022-05-30
Revert "Move to vscode-languageclient 8.0.1"

angelozerr:
This reverts commit ad42b24d32031e173465beb655c06f2fa8d3dc1d.


rgrunber (approved):
Thanks. For any reference on why we're doing this. See https://github.com/redhat-developer/vscode-java/pull/2377

Theia does not support this yet.

PR #718, opened 2022-06-08
Add button to double heap size on OOM

### LINKED ISSUE #700: Add option to increase heap maximum in OOM notification
I need to give the thing more memory.
It says "go to xml.server.vmargs and add -Xmx1G.
I search for xml.server.vmargs in settings and press Edit in settings.json
image
thats what the file looks like
image
where do i put the " -Xmx1G."

datho7561:
Copied from vscode-java.

Closes #700


datho7561:
I've tried testing the case where the user sets the vmargs to `"   "`, so that no -Xmx is set within them. From what I can tell, this means that the Java VM takes 1/4 of the available physical memory as the Xmx. I am having trouble getting lemminx to crash under this circumstance, since most of my computers have decently large amounts of physical memory, and VS Code is not good at opening large files (It crashes and freezes before lemminx does). I'll try again in a memory constrained VM.

datho7561:
Even on a memory constrained VM, I couldn't get the OOM message to appear when using empty jvm args.

rgrunber (approved):
Ok, so the subscription of the JDK Config changes happens before the LS starts interacting with the file.

Change looks good to me. Feel free to merge.

PR #721, opened 2022-06-20
Fix native image build on Windows builders.

rgrunber:
- Use git-clone over 'ssh' instead of 'https'
- Remove unnecessary setting of TLS for git-clone operation

### DIFF
--- a/NativeImage.jenkins
+++ b/NativeImage.jenkins
@@ -38,8 +38,10 @@ pipeline {
             label "win10"
           }
           steps {
-            powershell "if (Test-Path lemminx-win32.exe) { Remove-Item lemminx-win32.exe }"
-            powershell "Remove-Item -Recurse -Force lemminx"
+            powershell """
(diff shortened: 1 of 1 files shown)


NativeImage.jenkins
```diff
@@ -38,8 +38,11 @@ pipeline {
             label "win10"
           }
           steps {
-            powershell "if (Test-Path lemminx-win32.exe) { Remove-Item lemminx-win32.exe }"
-            powershell "Remove-Item -Recurse -Force lemminx"
+            powershell """
+              if (Test-Path lemminx-win32.exe) { Remove-Item lemminx-win32.exe }
+              Remove-Item -Recurse -Force lemminx
+              [Net.ServicePointManager]::SecurityProtocol
```
datho7561, line 44: This just outputs the https security protocol in use by powershell, right? If so, do we need this here?

rgrunber: I can remove it. When I considered setting to Tls13, I decided it might be good to know what the builder is set as prior to changing it.

datho7561 (approved):

PR #725, opened 2022-06-28
Add support for pre-releases.

rgrunber:
- VS Code Engine to 1.63.0
- Upgrade build jobs to use NodeJS 14
- Add prepare_pre_release gulp task

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -34,6 +34,22 @@ gulp.task('build_server_with_binary', function(done) {
 	done();
 });
 
+gulp.task('prepare_pre_release', function(done) {
+    const json = JSON.parse(fse.readFileSync("./package.json").toString());
+    const stableVersion = json.version.match(/(\d+)\.(\d+)\.(\d+)/);
+    const major = stableVersion[1];
+    const minor = stableVersion[2];
+    const date = new Date();
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
+    const hours = date.getHours();
+    const patch = `${date.getFullYear()}${prependZero(month)}${prependZero(day)}${prependZero(hours)}`;
+    const insiderPackageJson = Object.assign(json, {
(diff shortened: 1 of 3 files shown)


rgrunber:
This needs a *careful* rebase because one of the improvements I discovered in here to properly release platform-specific and generic vsixs needed to go directly into the release.

gulpfile.js
```diff
@@ -34,6 +34,22 @@ gulp.task('build_server_with_binary', function(done) {
...
+    const stableVersion = json.version.match(/(\d+)\.(\d+)\.(\d+)/);
+    const major = stableVersion[1];
+    const minor = stableVersion[2];
+    const date = new Date();
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
+    const hours = date.getHours();
+    const patch = `${date.getFullYear()}${prependZero(month)}${prependZero(day)}${prependZero(hours)}`;
+    const insiderPackageJson = Object.assign(json, {
+        version: `${major}.${minor}.${patch}`,
```
datho7561, line 48: I think it might be a good idea to preserve the patch number from the `package.json`, and attach the build time info afterwards with a `+`. eg. `0.21.1+2022070614`  See https://semver.org/#spec-item-10

datho7561 (changes requested):

gulpfile.js
datho7561: Roland informed me this isn't supported https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions

datho7561 (approved):

gulpfile.js
rgrunber: I think we should follow https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions since for now it seems they support strictly `major.minor.patch`

Issue #735, opened 2022-07-13
The formatter doesn't seem to use the VSCode proxy for downloading xsd files


halloleo:
When the extension formats a pom.xml (which contains an XML Schema) VS Code issues the problem "_Error while downloading 'http://maven.apache.org/xsd/maven-4.0.0.xsd'..._".

I have defined a proxy in VS Code's settings, but I assume the extension does not obey/use this proxy.

Is there another way to tell the formatter to use a proxy?

fbricon:
Please see if https://github.com/redhat-developer/vscode-xml/blob/master/docs/Proxy.md helps

halloleo:
Thanks @fbricon ! 

I have set `http.proxy` in vscode's settings.  I don't need authorisation.

I left `http.proxySupport` alone, showing "override" in the dialog.

This is from my understanding exactly what the page you link to says - but it doesn't work.

PS: On the page is the strange sentence: "*If you set [http.proxySupport] to override, VS Code will attempt to rewrite the requests from extensions to go through the proxy that is configured for VS Code. **However, this method sometimes does not work correctly.***" - In my case it doesn't seem to work correctly indeed, so what can I do?

fbricon:
@datho7561 can you please investigate?

datho7561:
Hello! I've tried on my machine with [this proxy](https://hub.docker.com/r/ubuntu/squid), and everything appears to be working. 

Which version of the extension are you using? Which version of VS Code (/Codium/Theia/?) are you using?

One thing that might be related is that http://maven.apache.org/xsd/maven-4.0.0.xsd redirects to the https version of the website; maybe something goes wrong when trying to handle this redirect? I'll research some more.

halloleo:
Just FYI: When I do 

    wget http://maven.apache.org/xsd/maven-4.0.0.xsd

with the px-proxy entered in my `.wgetrc`, hen I can download the schema without problems.

datho7561:
I was able to get a proper setup where my machine needs to use the proxy to access the http/https, and I am able to reproduce the issue.

datho7561:
I figured out the issue and am working on the PR. In the mean time, here is a workaround: https://github.com/redhat-developer/vscode-xml/blob/master/docs/Preferences.md#server-vm-arguments. My PR will just make sure that these parameters are set for you if you have set `http.proxy` in the VS Code settings.

halloleo:
Thanks that you looked into this! :)

[...]

PR #738, opened 2022-07-14
Also use VS Code proxy config for HTTPS on language server

datho7561:
One part of #735

### DIFF
--- a/docs/Preferences.md
+++ b/docs/Preferences.md
@@ -19,7 +19,7 @@
 
   Can be accessed through `xml.server.vmargs`.
 
-  Setting up proxy:
+### Setting up Proxy
 
   A proxy can be defined in the following way if there is a system proxy:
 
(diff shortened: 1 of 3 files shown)


docs/Preferences.md
```diff
@@ -19,7 +19,7 @@
 
   Can be accessed through `xml.server.vmargs`.
 
-  Setting up proxy:
+  eg. Setting up proxy:
```
AlexXuChen, line 22: Could we make this a header instead? Then it would look something like:

Can be accessed through `xml.server.vmargs`.


### Setting up proxy:

A proxy can be defined in the following way if there is a system proxy:

docs/Proxy.md
```diff
@@ -4,10 +4,11 @@ This page explains how to get vscode-xml to work through a proxy.
 
 ## Prerequisites
 
-If you want to use the binary language server,
-then you will need to make sure that your proxy allows connections to jboss.org
-since the binary language server is downloaded from that website.
-You will also need to make sure that you can access any online XML schemas that you want to work with through the proxy.
+You need to make sure that you can access any online XML schemas that you want to work with through the proxy.
+Also, if are using open-vsx.org instead if Visual Studio Marketplace for extensions,
```
AlexXuChen, line 8: "...instead of Visual..."

AlexXuChen: I think you can remove the comma in "...for extensions, and..."

AlexXuChen (changes requested):

rgrunber (approved):
Looks good to me.

I also noticed that phrasing :

`The proxy setting to use. If not set, will be inherited from the http_proxy and https_proxy environment variables.` . I'm guessing we don't support extracting those environment variables on the client side ?

datho7561:
Thanks for the reviews!

PR #743, opened 2022-07-20
Added documentation for `XML: Restart Language Server`

AlexXuChen:
Added documentation for `XML: Restart Language Server`.

As per https://github.com/redhat-developer/vscode-xml/pull/551#pullrequestreview-1045491718

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -25,3 +25,7 @@ When the [Server Cache Path](Preferences.md#server-cache-path) is activated, the
 This command re-triggers the [XML Validation](Validation.md#xml-validation) for the all opened XML files.
 
 When the [Server Cache Path](Preferences.md#server-cache-path) is activated, the command clears the remote grammar cache and revalidates all opened files.
+
+## Restart XML Language Server
+
+This command restarts the XML language server.
--- a/package.json
+++ b/package.json
@@ -616,7 +616,7 @@
       },
       {
         "command": "xml.restart.language.server",
-        "title": "Restart Language Server",
+        "title": "Restart XML Language Server",
         "category": "XML"
       }
     ],


docs/Commands.md
```diff
@@ -25,3 +25,7 @@ When the [Server Cache Path](Preferences.md#server-cache-path) is activated, the
 This command re-triggers the [XML Validation](Validation.md#xml-validation) for the all opened XML files.
 
 When the [Server Cache Path](Preferences.md#server-cache-path) is activated, the command clears the remote grammar cache and revalidates all opened files.
+
+## Restart XML Language Server connection
+
+This command triggers `XML: Restart Language Server`, which restarts the connection from the vscode client and the XML language server.
```
datho7561, line 31: Does it just restart the connection, or does LemMinX restart?

AlexXuChen: Just the connection, it calls LanguageClient.start() and LanguageClient.stop()

AlexXuChen: In hind sight, using the name `XML: Restart Language Server Connection` might have been better

datho7561: I think that it actually restarts the server. When I run `watch jps`, then issue the command, the PID of the server changes. I think it might be a better idea to say "restart the server", in the documents

rgrunber: I would keep the description simple :

This command restarts the XML language server.

I don't think we need to go into any detail on how it does it (stop/start) or to specify that we do it vita the vscode language client API. As far as the user is concerned. the current language server process is shut down, and a new one replaces it.

docs/Commands.md
```diff
@@ -25,3 +25,7 @@ When the [Server Cache Path](Preferences.md#server-cache-path) is activated, the
 This command re-triggers the [XML Validation](Validation.md#xml-validation) for the all opened XML files.
 
 When the [Server Cache Path](Preferences.md#server-cache-path) is activated, the command clears the remote grammar cache and revalidates all opened files.
+
+## Restart XML Language Server connection
+
+This command triggers `XML: Restart Language Server`.
```
rgrunber, line 31: - I would make the title "Restart XML Language Server"
- The body can just be "This command restarts the XML language server"

datho7561: @AlexXuChen hey, do you have time to take a look at this today?

AlexXuChen: Fixed

rgrunber (approved):

datho7561 (approved):

PR #744, opened 2022-07-20
Replace tslint with eslint

### LINKED ISSUE #438: Set up linting for vscode-xml
It might be worthwhile to set up a linter, such as [ESLint](https://eslint.org/) for vscode-xml.

datho7561:
Configure VS Code integration for eslint. Add a GitHub Actions workflow that runs the TypeScript compiler and eslint.

Closes #438


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,17 @@
+name: lint
```
rgrunber, line 1: Agreed. We should run this as part of pull requests / commits.

src/server/requirements.ts
```diff
@@ -16,12 +17,6 @@ export interface RequirementsData {
...
 
-interface ErrorData {
-  message: string;
-  label: string;
-  openUrl: Uri;
-  replaceClose: boolean;
-}
 /**
  * Resolves the requirements needed to run the extension.
  * Returns a promise that will resolve to a RequirementsData if
```
rgrunber, line 22: Just below here it mentions ErrorData. Can this just be removed now ?

src/server/requirements.ts
```diff
@@ -5,9 +5,10 @@ import * as path from 'path';
 import { ConfigurationTarget, env, ExtensionContext, Uri, window, workspace } from 'vscode';
 import { getJavaagentFlag, getJavaConfiguration, getKey, getXMLConfiguration, IS_WORKSPACE_JDK_ALLOWED, IS_WORKSPACE_JDK_XML_ALLOWED, IS_WORKSPACE_VMARGS_XML_ALLOWED, xmlServerVmargs } from '../settings/settings';
 
-const pathExists = require('path-exists');
-const expandHomeDir = require('expand-home-dir');
-const findJavaHome = require('find-java-home');
+import pathExists = require('path-exists');
+import expandHomeDir = require('expand-home-dir');
+import findJavaHome = require('find-java-home');
```
rgrunber, line 10: You'll need to remove this when you rebase since we replaced `find-java-home` with `jdk-utils`

rgrunber (commented):
When I rebase the change, and run `npm run compile`, I get :

```
ERROR in ./src/server/requirements.ts 72:33
Module parse failed: The keyword 'yield' is reserved (72:33)
File was processed with these loaders:
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
|             }
|             //No settings, let's try to detect as last resort.
>             const javaRuntimes = yield (0, jdk_utils_1.findRuntimes)({ withVersion: true, withTags: true });
|             if (javaRuntimes.length) {
|                 sortJdksBySource(javaRuntimes);
 @ ./src/extension.ts 34:21-53

ERROR in /home/rgrunber/git/vscode-xml/src/server/requirements.ts
./src/server/requirements.ts 63:25-30
[tsl] ERROR in /home/rgrunber/git/vscode-xml/src/server/requirements.ts(63,26)
      TS1308: 'await' expressions are only allowed within async functions and at the top levels of modules.
```

src/server/requirements.ts
datho7561: ErrorData wasn't being used to annotate anything, so I annotated the object that this was supposed to be used for. However, we should remove a lot of this code at some point, since we never do anything with the ErrorData object we make.

.github/workflows/lint.yaml
datho7561: Should be set up already I think: https://github.com/redhat-developer/vscode-xml/actions/runs/2861015002

src/server/requirements.ts
datho7561: See https://github.com/redhat-developer/vscode-xml/issues/761

rgrunber (approved):
Change works well for me.

PR #756, opened 2022-08-02
Migrate from find-java-home to jdk-utils

### LINKED ISSUE #627: Consider migrating from find-java-home to jdk-utils
See https://github.com/redhat-developer/vscode-java/pull/2246 for reference point.
https://www.npmjs.com/package/jdk-utils

JessicaJHee:
Fixes #627 

### DIFF
--- a/src/server/requirements.ts
+++ b/src/server/requirements.ts
@@ -4,10 +4,10 @@ import * as cp from 'child_process';
 import * as path from 'path';
 import { ConfigurationTarget, env, ExtensionContext, Uri, window, workspace } from 'vscode';
 import { getJavaagentFlag, getJavaConfiguration, getKey, getXMLConfiguration, IS_WORKSPACE_JDK_ALLOWED, IS_WORKSPACE_JDK_XML_ALLOWED, IS_WORKSPACE_VMARGS_XML_ALLOWED, xmlServerVmargs } from '../settings/settings';
+import { findRuntimes, IJavaRuntime, getSources } from 'jdk-utils';
 
 const pathExists = require('path-exists');
 const expandHomeDir = require('expand-home-dir');
-const findJavaHome = require('find-java-home');
 const isWindows = process.platform.indexOf('win') === 0;
 const JAVA_FILENAME = 'java' + (isWindows ? '.exe' : '');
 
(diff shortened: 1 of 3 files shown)


src/server/requirements.ts
```diff
@@ -66,14 +66,14 @@ function checkJavaRuntime(context: ExtensionContext): Promise<string> {
...
     //No settings, let's try to detect as last resort.
-    findJavaHome({ allowJre: true }, function (err, home) {
-      if (err) {
-        openJDKDownload(reject, 'Java runtime could not be located.');
-      }
-      else {
-        resolve(home);
-      }
-    });
+    const javaRuntimes = await findRuntimes({ checkJavac: true, withVersion: true, withTags: true });
```
datho7561, line 69: I don't think that we need to check for `javac` here, since we just need a JRE (to run java software) not a full JDK (to run and develop/compile java software)

datho7561 (changes requested):
You should also be able to get rid of lines 48 thru 55 in `requirements.ts`, since jdk_utils can find a JDK if JAVA_HOME or JDK_HOME is set.

datho7561 (changes requested):
I tried it out, and its working like a charm! I still think the "no java" message could be improved

datho7561 (approved):
Looks good to me! Thanks, Jessica!

rgrunber:
Just a small note here. I ran `npm install` freshly and got :

```diff
$ git diff
diff --git a/package-lock.json b/package-lock.json
index f0a204c..e635cac 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -4953,11 +4953,6 @@
         "source-map": "^0.6.1"
       }
     },
-    "which": {
-      "version": "1.0.9",
-      "resolved": "https://registry.npmjs.org/which/-/which-1.0.9.tgz",
-      "integrity": "sha1-RgwdoPgQED0DIam2M6+eV15kSG8="
-    },
     "which-module": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/which-module/-/which-module-1.0.0.tgz",
@@ -4970,11 +4965,6 @@
       "integrity": "sha512-JcKqAHLPxcdb9KM49dufGXn2x3ssnfjbcaQdLlfZsL9rH9wgDQjUtDxbo8NE0F6SFvydeu1VhZe7hZuHsB2/pw==",
       "dev": true
     },
-    "winreg": {
-      "version": "1.2.4",
-      "resolved": "https://registry.npmjs.org/winreg/-/winreg-1.2.4.tgz",
-      "integrity": "sha1-ugZWKbepJRMOFXeRCM9UCZDpjRs="
-    },
     "wrap-ansi": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/wrap-ansi/-/wrap-ansi-2.1.0.tgz",

```

So a change somewhere resulted in us not needing these dependencies. Have a look at https://github.com/redhat-developer/vscode-xml/pull/756/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519 definition of find-java-home. It's likely the module was the only one requiring these, so if you just ran an `npm install` after making the replacement in the package.json, it would have done that for you. In general, you want to avoid manually modifying the package-lock.json.

PR #765, opened 2022-08-17
Read proxy config from the environment

### LINKED ISSUE #764: Obey https_proxy settings
I have http_proxy and https_proxy set yet the extension is most likely failing to use them.

Please provide better behaviour than jvm defaults which are horrible.

See #134 for details.

datho7561:
Reads the environment variable `http_proxy` as the proxy config if the setting `"http.proxy"` is not set.

Closes #764

### DIFF
--- a/src/settings/proxySettings.ts
+++ b/src/settings/proxySettings.ts
@@ -124,8 +124,9 @@ export function getProxySettingsAsJVMArgs(proxySettings: ProxySettings): string 
  * @param proxySettings the proxy settings to convert into environment variables
  * @returns the proxy settings as environment variables for LemMinX
  */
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
If you can confirm it works, feel free to merge. It seems like we briefly discussed it at https://github.com/redhat-developer/vscode-xml/pull/738#pullrequestreview-1040356617 and were fine with leaving it out then.

**Update** : Is it worth also checking `https.proxy` ?

datho7561:
> It seems like we briefly discussed it at [#738 (review)](https://github.com/redhat-developer/vscode-xml/pull/738#pullrequestreview-1040356617)

Oh no! I should be more careful about these things in the future.

Issue #766, opened 2022-08-18
Update badges in README


datho7561:
It would be nice to update the badges at the top of the README to use shields.io. See https://github.com/redhat-developer/vscode-quarkus/pull/530 for an example.

Here is what I have in mind:
* Fix VS Code marketplace version badge (current is broken)
* Add a badge for the ESLint GitHub Action
* Keep the Gitter badge, but update it to use shields.io

PR #767, opened 2022-08-22
Installs badge

angelozerr:
Installs badge


angelozerr:
This PR adds installs badge:

image

datho7561 (approved):
I think it'd be nice to also add a badge for eslint, (see my issue #766), but we could also do that later.

angelozerr:
> I think it'd be nice to also add a badge for eslint, (see my issue https://github.com/redhat-developer/vscode-xml/issues/766), but we could also do that later.

Yes sure, but let's do that in an another PR.

PR #768, opened 2022-08-23
Update badges

### LINKED ISSUE #766: Update badges in README

datho7561:
Closes #766

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 # XML Language Support by Red Hat
 
-[![Join the chat at https://gitter.im/redhat-developer/vscode-xml](https://badges.gitter.im/redhat-developer/vscode-xml.svg)](https://gitter.im/redhat-developer/vscode-xml?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Marketplace Version](https://vsmarketplacebadge.apphb.com/version/redhat.vscode-xml.svg "Current Release")](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml)
-[![Installs](https://img.shields.io/visual-studio-marketplace/i/redhat.vscode-xml)](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,9 @@
...
-[![Installs](https://img.shields.io/visual-studio-marketplace/i/redhat.vscode-xml)](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml)
+[![Join the chat at https://gitter.im/redhat-developer/vscode-xml](https://img.shields.io/gitter/room/redhat-developer/vscode-xml?color=ED0060&style=for-the-badge&logo=gitter)](https://gitter.im/redhat-developer/vscode-xml?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Marketplace Version](https://img.shields.io/visual-studio-marketplace/v/redhat.vscode-xml?color=informational&logo=visualstudiocode&style=for-the-badge)](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml)
+[![Installs](https://img.shields.io/visual-studio-marketplace/i/redhat.vscode-xml?logo=visualstudiocode&color=informational&style=for-the-badge)](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml)
+[![eslint github action](https://img.shields.io/github/workflow/status/redhat-developer/vscode-xml/lint?label=eslint&logo=eslint&style=for-the-badge)](https://github.com/redhat-developer/vscode-xml/actions/workflows/lint.yaml?query=branch%3Amain)
```
rgrunber, line 6: Just wondering what the `text inside the ![....]` does, because it doesn't correspond to the labels. Also for the marketplace I would make the label `VS Marketplace` so it matches what we do in other places.

rgrunber: Ah, it's supposed to be alt text for the image, although I don't see it :\ . Ok then just fix the above label.

datho7561: One way to test the alt text could be rendering the markdown while not connected to the internet. I'll take a look into updating the marketplace alt text

datho7561: Okay, should be updated properly and rebased

rgrunber (approved):

PR #779, opened 2022-09-20
Fix problem matcher

datho7561:


datho7561:
Okay, I've updated the infrastructure logging level to the same as in vscode-quarkus and vscode-microprofile

rgrunber (approved):
Works for me now. The only way to reproduce is to run immediately after a freshly opened workspace of vscode-xml. Otherwise, any changes seem to make it into the extension child instance being run.

PR #782, opened 2022-10-06
Report only XML syntax error for *.exsd files

angelozerr:
Report only XML syntax error for *.exsd files

### DIFF
--- a/docs/Validation.md
+++ b/docs/Validation.md
@@ -544,10 +544,15 @@ By default, vscode-xml uses this default validation filter:
 
 ```json
 "xml.validation.filters": [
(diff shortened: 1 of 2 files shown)


angelozerr:
> I think that this setting needs some documentation, especially since it doesn't ignore errors coming from <?xml-model processing instructions yet

It is a bug,so we should create an issue on LemMinx side

datho7561:
@angelozerr do you think you could add some documentation for this setting in the PR?

angelozerr:
> @angelozerr do you think you could add some documentation for this setting in the PR?

Sorry to do that in late, fixed.

PR #788, opened 2022-10-14
Fix syntax coloration with element/attribute which can contains '-'

angelozerr:
Fix syntax coloration with element/attribute which can contains '-'

This PR is a smal fix to support `-` for attribute/element name.

Given this rnc file:

```
element addressBook {
  element my-card {
    attribute my-id { text },
    element name { text },
    element email { text }
  }*
}
```

With the master we have:

image

With the PR element/attribute is colorized correctly (not blank)

image


datho7561 (approved):

angelozerr:
Thanks @datho7561 !

PR #789, opened 2022-10-14
Make sure that 'vsce' is installed in nodes that use it.

rgrunber:
The 'vsce package' and 'vsce publish' stages happen in separate node (machine) environments. While both are 'rhel8', there is no guarantee the same node will be used in both cases, so we should explicitly ensure the tool is installed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -104,6 +104,8 @@ node('rhel8'){
 }
 
 node('rhel8'){
+	sh 'npm install -g "vsce"'
+	sh 'npm install -g "ovsx"'
 	if(publishToMarketPlace.equals('true') || publishPreRelease.equals('true')){
 
 		if (publishToMarketPlace.equals('true')) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -127,6 +127,7 @@ node('rhel8'){
 
 		if (publishToMarketPlace.equals('true')) {
 			// Open-VSX Marketplace does not support pre-release
+			sh 'npm install -g "vsce"'
```
datho7561, line 130: Why not install this at the beginning of the `node` block (eg. like 106/108)? It looks like `vsce` is run on line line 122, before this install is run

datho7561 (changes requested):

Jenkinsfile
rgrunber: ah whoops. this currently does nothing.

rgrunber: Done. Should be at the top of the final 'node' declaration that publishes..

datho7561 (approved):
Looks good. Thanks, Roland!

PR #798, opened 2022-10-27
Update joinContentLines documentation with new behavior

JessicaJHee:

### DIFF
--- a/docs/Formatting.md
+++ b/docs/Formatting.md
@@ -305,28 +305,66 @@ If this value is set to 0, then all blank lines will be removed during formattin
 
 ### xml.format.joinContentLines
 
-Normalize the whitespace of content inside an element. Newlines and excess whitespace are removed. Default is `false`.
+Set to `true` to normalize the whitespace of content inside an element. Newlines and excess whitespace are removed. Default is `false`.
 
-For example, the following document doesn't change if it is set to `false`:
+When `xml.format.joinContentLines` is set to `false`, the following edits will be made:
+
+* text following a line separator will be appropriately indented
+* spaces between text in the same line will be normalized
+* any exisiting new lines will be treated with respect to the [`xml.format.preservedNewlines`](#xmlformatpreservednewlines) setting
+
+For example, before formatting:
 
   ```xml
   <?xml version='1.0' encoding='UTF-8'?>
   <root>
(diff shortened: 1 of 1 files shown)


docs/Formatting.md
```diff
@@ -305,18 +305,40 @@ If this value is set to 0, then all blank lines will be removed during formattin
...
-Normalize the whitespace of content inside an element. Newlines and excess whitespace are removed. Default is `false`.
+Set to `true` to normalize the whitespace of content inside an element. Newlines and excess whitespace are removed. Default is `false`.
 
-For example, the following document doesn't change if it is set to `false`:
+When joinContentLines is set to `false`, the following edits will be made:
+
+* text following a line separator will be appropriately indented
+* spaces between text in the same line will be normalized
+* any exisiting new lines will be treated with respect to the [`xml.format.preservedNewlines`](#xmlformatpreservednewlines) setting
+
```
angelozerr, line 315: I think you should mention that in the following samples, preservedNewlines = 2 (default value), because we have the impression that it doesn't remove empty lines.

It could be interesting to add the result too with preservedNewlines=0 to show that it removes empty line

angelozerr: We should mention too I think in preservedNewlines section that it works only with tag element, not with just text. When I test that I though preservedNewlines was broken. I think we could add a new settings like `xml.format.preservedNewlinesInText` but it is an another issue.

JessicaJHee: Could you please post an example xml where preservedNewlines doesn't work?

angelozerr: Here a sample with only text:

```xml
<foo>


	gggggggggggggggggg
	ooooooooooooooooooo


</foo>
```

after formatting, we have:

`<foo> gggggggggggggggggg ooooooooooooooooooo </foo>`

Otherwise if the XML use tag elements:

```xml
<foo>


	<gggggggggggggggggg />
	<ooooooooooooooooooo />


</foo>
```

it works perfectly.

JessicaJHee: Hmm that's strange, it works for me

Peek 2022-10-27 16-16

angelozerr (approved):

angelozerr:
Thanks @JessicaJHee !

PR #806, opened 2022-11-15
no svg validation if jock.svg is also installed

### LINKED ISSUE #805: no svg validation if `jock.svg` is also installed
If I have a group and remove the end tag `</g>`, I get no error, so the validation does not work.

angelozerr:
no svg validation if jock.svg is also installed

Fixes #805

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "onLanguage:xml",
     "onLanguage:xsl",
     "onLanguage:dtd",
+    "onLanguage:svg",
     "onCommand:xml.open.docs",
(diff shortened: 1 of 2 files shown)


angelozerr:
Here a demo 

SVGSupport

You can see syntax error validation, but not SVG validation because it is not linked to a SVG DTD.

The completion that you see in the demo comes from the jock.svg vscode extension. To do that we need to define a XML catalog.

datho7561 (approved):
This works for me! Thanks, Angelo!

PR #807, opened 2022-11-15
Provide the capability to bind RelaxNG schema

angelozerr:
Provide the capability to bind RelaxNG schema

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -159,6 +159,12 @@ for (const label of bindingTypes.keys()) {
   bindingTypeOptions.push({ "label": label });
 }
 
+const bindingTypeOptionsForRelaxNG: QuickPickItem[] = [];
+for (const entry of bindingTypes.entries()) {
+  if (entry[1] !== 'standard') {
+    bindingTypeOptionsForRelaxNG.push({ "label": entry[0] });
+  }
+}
 /**
  * The function passed to context subscriptions for grammar association
  *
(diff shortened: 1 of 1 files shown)


src/commands/registerCommands.ts
```diff
@@ -178,9 +178,9 @@ async function grammarAssociationCommand(documentURI: Uri, languageClient: Langu
   // Open a dialog to select the XSD, DTD to bind.
   const options: OpenDialogOptions = {
     canSelectMany: false,
-    openLabel: 'Select XSD or DTD file',
+    openLabel: 'Select XSD, DTD, RNG or RNC file',
```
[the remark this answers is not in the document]
angelozerr: Good catch!

I wonder if it should be better to select the file before the strategy.

If you are newbie with relaxng you dont know that you will need to select xml-model.

I think too that we should provide the capability to fill an URL too and continue to select a file too. 

My idea is to have at first à choice local / remote grammar.

If you select local it opens the file dialog and after sélection of file it opens the binding strategies by hiding standard If rng rnc file is selected.

If you select remote it opens à simple file where you can file the URL and after it show stratgies.

What do you think about this idea?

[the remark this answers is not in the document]
datho7561: I think that it makes sense to select the grammar file before the binding strategy.

The thing with the remote grammar/typing in a URI sounds cool, but I think that fits better in a separate PR.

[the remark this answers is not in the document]
angelozerr: > I think that it makes sense to select the grammar file before the binding strategy.

fixed

> The thing with the remote grammar/typing in a URI sounds cool, but I think that fits better in a separate PR.

I created the issue https://github.com/redhat-developer/vscode-xml/issues/808

src/commands/registerCommands.ts
```diff
@@ -159,6 +159,12 @@ for (const label of bindingTypes.keys()) {
   bindingTypeOptions.push({ "label": label });
 }
 
+const bindingTypeOptionsForRelaxNG: QuickPickItem[] = [];
+for (const entry of bindingTypes.entries()) {
+  if (entry[1] != 'standard') {
```
datho7561, line 164: it's generally a good idea to use `!==` instead of `!=` (it won't make a difference in the logic in this case, though).

(apparently the eslint rule to check this, `"eqeqeq"`, is off by default, which is an odd decision)

src/commands/registerCommands.ts
```diff
@@ -168,38 +174,41 @@ for (const label of bindingTypes.keys()) {
...
+  const pickedBindingTypeOption = await window.showQuickPick(bindingTypesQuickPick, { placeHolder: "Binding type" });
+  if (!pickedBindingTypeOption) {
+    return;
+  }
+  const bindingType = bindingTypes.get(pickedBindingTypeOption.label);
+
+  // The XSD, DTD has been selected, get the proper syntax for binding this grammar file in the XML document.
+  try {
+    const currentFile = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ? window.activeTextEditor.document : undefined;
+    if (bindingType == 'fileAssociation') {
```
datho7561, line 203: Same comment about `==` here

src/commands/registerCommands.ts
angelozerr: I missed that, thanks for reporting that.

fixed

src/commands/registerCommands.ts
angelozerr: fixed

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #809, opened 2022-11-17
Fill remote URL with "Bind to grammar / schema"

### LINKED ISSUE #808: Fill remote URL with "Bind to grammar / schema"
Today whenwe click on `Bind to grammar / schema` codelens, we can only select local file.

It should be nice to provide the capability to fill a remote URL too.

See https://github.com/redhat-developer/vscode-xml/pull/807#discussion_r1023172808

angelozerr:
Fill remote URL with "Bind to grammar / schema"

Fixes #808


angelozerr:
In my case, when I want to bind an XML file with a given URL, I copy in the clipboard the URL.

When `Bind schema / grammar` is clicked and remote is selected, the field is prefilled with the content of the clipboard (if it starts with http). It avoid pasting the URL.

datho7561 (approved):
Looks good! Thanks Angelo!

PR #811, opened 2022-11-17
Add MEI, TEI, SVG as keywords

angelozerr:
As vscode-xml can supports MEI, TEI and SVG, it should be nice to add those keywords when user will search extension:

 * `MEI` : https://music-encoding.org/
 * `TEI` : https://tei-c.org/
 * `SVG`: https://developer.mozilla.org/en-US/docs/Web/SVG

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,10 @@
     "dtd",
     "RelaxNG",
     "rng",
-    "rnc"
+    "rnc",
+    "TEI",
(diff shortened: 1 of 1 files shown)


datho7561 (approved):
I don't think it takes effect until it's released on the marketplace, so I can't test it, but it sounds like a great idea! Thanks, Angelo!

PR #813, opened 2022-11-18
Wrap selection in XML element

### LINKED ISSUE #794: Wrap selection in XML element
Most XML editors have a feature to wrap the current selection in an XML element, providing completion to suggest valid tags at that point in a file.

Within VS Code, the built-in 'Emmet: Wrap with Abbreviation' function provides part of this, but it is not XML-aware and does not give a list of the allowed tags in the current context. It would thus be brilliant if the XML extension could provide this.

There are at least two ways to implement this:

1. Inline: TextMate allows the user to make a selection, then type <kbd>Ctrl</kbd> + <kbd><</kbd> to insert open/close tags around it, balancing the element name as the user types. The VS Code XML plugin could use this approach to immediately provide a list of options identical to the current functionality when typing <kbd><</kbd>.
2. With a dialogue box: Similar to the existing Emmet function but with autocompletion. The Oxygen XML Editor, for example, provides such a dialogue with the shortcut <kbd>Ctrl</kbd> + <kbd>E</kbd>. The [Scholarly XML](https://marketplace.visualstudio.com/items?itemName=raffazizzi.sxml) extension for VS Code has an implementation of this, but it conflicts with this XML plugin.

Many thanks for your work on this project and considering this request!

angelozerr:
Wrap selection in XML element

Fixes #794


angelozerr:
I created very quickly wrap element feature, but it needs some fix on LemMinx side and require to provide a command on LemMinx side which provide the proper before start tag element / after end tag element position.

Once we will have this command, we could provide a proper wrapped element and even another feature like comment to comment a selected element.

Here a demo:

WrapElementDraftDemo

Here the limitation of this PR:

 * you need to select the whole element to know the start / end position. If you select a part of the element, it will insert the end tag element inside the existing element and will break the XML. To fix that, we need a command on LemMinx side which give the proper position for start / tag of the selected element
 
WrapElementBug2

 * the completion appears, but if you apply it, it generates end element just after the start tag element 
WrapElementBug1

otichy:
This is great, thanks for starting the work on this!

As to the first limitation - imho, this is a feature rather than a bug - the function should wrap whatever is selected, be it a string/text, element or even a part of an element. If the selection leads to a malformed xml, that's the user's problem.

angelozerr:
> the completion appears, but if you apply it, it generates end element just after the start tag element

This problem will be fixed with the current PR https://github.com/eclipse/lemminx/pull/1374

> This is great, thanks for starting the work on this!

You are welcome!

> As to the first limitation - imho, this is a feature rather than a bug - the function should wrap whatever is selected, be it a string/text, element or even a part of an element. If the selection leads to a malformed xml, that's the user's problem.

You mean that we should not improve that? How Oxygen takes care of this problem?

My idea was to adjust selection to wrap the covered element, is not a good idea?  @otichy @adunning what is your suggestion about that?

otichy:
> You mean that we should not improve that? How Oxygen takes care of this problem?

oXygen does exactly what your current implementation does - it respects the span of the user's selection, even if the result of the wrapping action is invalid XML.
 
> My idea was to adjust selection to wrap the covered element, is not a good idea? @otichy @adunning what is your suggestion about that?

I have to say that I'd personally prefer if the user's selection was respected exactly. There are many occasions when temporarily invalid code is actually intentional and I'd find attempts to correct my selection for the purposes of wrapping more frustrating than helpful. But this could just be because I am used to how it works in oXygen.

angelozerr:
@otichy ok let's start with this simple implementation and we will see in the future if adjust can be a good idea.

What about if there are not selection? It will insert the start/end tag in the current cursor, is that?

otichy:
> What about if there are not selection? It will insert the start/end tag in the current cursor, is that?

Yes, exactly. oXygen surrounds the current cursor position with an empty tag.

Btw, were you able to solve the second limitation? That seems to be the only problem to me at the moment.

angelozerr:
> Yes, exactly. oXygen surrounds the current cursor position with an empty tag.

Thanks for you feedback!

> Btw, were you able to solve the second limitation? That seems to be the only problem to me at the moment.

Exactly, and I created a PR https://github.com/eclipse/lemminx/pull/1374 (which was not trivial to implement, so I hope I will not break something, but we have a lot of tests and I have written new tests again, so I hope we will have no trouble)

@datho7561 is reviewing this PR, I hope we could merge it this week, if everything works correctly. After that I will work again on this PR just to clean some piece of code.

Another question, I initialize the tag element with `<elt>`, but I think it should be better to use the first completion item (I don't know if it i shard to do that)

otichy:
> Another question, I initialize the tag element with `<elt>`, but I think it should be better to use the first completion item (I don't know if it i shard to do that)

Yes, the first completion option or nothing might be better to initialise with, rather than something arbitrary - or is `<elt>` one of the options?

otichy:
> > is  one of the options?
> 
> Not it is hard coded for the moment.

Is it necessary to initiate with an actual tag before the user selects the completion option?

datho7561:
A few things that came to mind:
- It would be nice if this supported multicursor. The Emmet extension manages this properly
- If you have just the cursor and use wrap element to create a new element, it would be nice if there was a tab stop at the end of the start tag, so that you can tab over and start adding the element.

JessicaJHee:
This change works really well for me! I wonder if it is a good idea to add this option to the menu that appears on right-click using this https://code.visualstudio.com/api/references/contribution-points#contributes.menus?

angelozerr:
This PTR requires https://github.com/eclipse/lemminx/pull/1389

src/commands/registerCommands.ts
```diff
@@ -366,3 +367,82 @@ function registerRestartLanguageServerCommand(context: ExtensionContext, languag
...
+  }
+
+  const startTag = result.start.newText;
+  const endTag = result.end.newText;
+  const startPos = languageClient.protocol2CodeConverter.asPosition(result.start.range.start);
+  const endPos = languageClient.protocol2CodeConverter.asPosition(result.end.range.start);
+  const pos = new Position(selection.start.line, selection.start.character + 1);
+
+  editor.insertSnippet(new SnippetString(endTag), endPos);
+  editor.insertSnippet(new SnippetString(startTag), startPos);
```
datho7561, line 436: If you apply these separately, then you have to undo twice to undo surround tag. I think it would be very helpful if we could avoid this.

angelozerr:
> It would be nice if this supported multicursor. The Emmet extension manages this properly

Could you give me a demo please.

> If you have just the cursor and use wrap element to create a new element, it would be nice if there was a tab stop at the end of the start tag, so that you can tab over and start adding the element.

After some investigation, it is possible to do that but with the new `SnippetTextEdit` from vscode (appears in september), but it means:

 * we need to bump the vscode dependency (I'm not sure that Theia supports it)
 * we will loose the documentation when completion will be opened (because in this case completion will be managed with snippet choice). @otichy @mkhl is it annoying to have completion without documentation?

> Other than that, it's working very well. I think that the decision not to bind it to a keybinding by default is a good idea.

Ok I removed keybinding.

datho7561:
Here's something that I can do with the Emmet surround with tag using multiline cursors:
emmet-surround-with-tag

datho7561:
Here's another example of how it could be useful:
emmet-surround-with-to-make-list

datho7561:
If you are wrapping existing content, such as:

```xml
<Burger>
  |Bacon|
</Burger>
```

Then I think that the tab stop at the end of the start tag isn't needed. So instead of `<elt>$1Bacon</elt>$2` you can just do `<elt>Bacon</elt>$1`

src/commands/registerCommands.ts
angelozerr: Sorry it was a POC, now I'm using SnippetTextEdit which works with WorkspaceEdit.

angelozerr:
> Here's something that I can do with the Emmet surround with tag using multiline cursors:

Thanks for the explanation, could we see that in an another PR?

datho7561:
> could we see that in an another PR?

Do you think it will be hard to implement?

angelozerr:
> Do you think it will be hard to implement?

I don't know, but I would like to finish this PR (I need to write documentation and clean code from LemMinX). If I have time I will investigate this feature.

angelozerr:
> Then I think that the tab stop at the end of the start tag isn't needed. So instead of <elt>$1Bacon</elt>$2 you can just do <elt>Bacon</elt>$1

Here a demo why I decided to do that:

SurroundTags

angelozerr:
@otichy here demo with suround with tags 

SurroundWithTags

with comments:

SurroundWithComments

angelozerr:
> then changing the @types/vscode in the package.json back.

You mean that package.json should keep the @types/vscode version like master?

otichy:
> * we will loose the documentation when completion will be opened (because in this case completion will be managed with snippet choice). @otichy @mkhl is it annoying to have completion without documentation?

It's not essential, but certainly useful - especially in cases of tagging with a complex schema (such as TEI) or in cases of large teams of encoders who may tag more consistently with documentation available in autocomplete.

angelozerr:
> It's not essential, but certainly useful - especially in cases of tagging with a complex schema (such as TEI) or in cases of large teams of encoders who may tag more consistently with documentation available in autocomplete.

Ok, we decide to manage surround with snippet to have a nice cursor management : the cursor is set at the end of the inserted end tag when completion is applied but we have no documentation.

We will see if we needto change that according user feedback (we could have a preferences to have documentation or cursor)

angelozerr:
I fixed the problem with XML command which appears on command palette, those commands appears only when XML Language Server is started. If you restart the seraver with the command, all commands should disappear until the XML language server is started.

docs/Refactor.md
```diff
@@ -0,0 +1,23 @@
...
+
+ * use command palette (`Ctrl+P`) and type `Surround`
+
+![Surround with Tags](images/Refactor/SurroundWithTags.gif)
+
+ * use contextual menu
+
+## Surround with Comments
+
+As `Surround with Tags (Wrap)`, you can comments a selected XML content:
```
datho7561, line 15: *Similar to* `Surround with Tags (Wrap)`, you can comment~~s~~ *out* the selected XML content:

docs/Refactor.md
```diff
@@ -0,0 +1,23 @@
...
+
+![Surround with Tags](images/Refactor/SurroundWithComments.gif)
+
+## Surround with CDATA
+
+As `Surround with Tags (Wrap)`, you can surround with CDATA a selected XML content:
```
datho7561, line 21: Similar to`Surround with Tags (Wrap)`, you can surround the selected XML content with CDATA

src/commands/serverCommandConstants.ts
```diff
@@ -24,9 +24,14 @@ export const ASSOCIATE_GRAMMAR_INSERT = "xml.associate.grammar.insert";
...
 
 /**
  * Command to check if a given file pattern matches any file on the workspace
  */
-export const CHECK_FILE_PATTERN = "xml.check.file.pattern"
\ No newline at end of file
+export const CHECK_FILE_PATTERN = "xml.check.file.pattern";
+
+/**
+ * Command to ...
```
datho7561, line 35: Finish this doc

src/client/xmlClient.ts
```diff
@@ -14,6 +14,17 @@ import * as Telemetry from '../telemetry';
...
 
+export const XML_SUPPORTED_LANGUAGE_IDS = ['xml', 'xsl', 'dtd', 'svg'];
+
+const XML_DOCUMENT_SELECTOR: DocumentSelector =
+  XML_SUPPORTED_LANGUAGE_IDS.map(function (langId) {
+    return { scheme: 'file', language: langId }
+  }).concat(
+    XML_SUPPORTED_LANGUAGE_IDS.map(function (langId) {
+      return { scheme: 'untitled', language: langId }
+    }));
```
datho7561, line 25: I think you should be able to make this a bit more compact and easy to read using the arrow syntax eg.:

```typescript
.map(langId => { scheme: 'file', language: langId })
```

package.json
```diff
@@ -684,17 +684,69 @@
...
+          "command": "xml.validation.all.files",
+          "when": "XMLLSReady"
         },
         {
           "command": "xml.command.bind.grammar",
-          "when": "resourceFilename =~ /xml/ && editorIsOpen"
+          "when": "resourceFilename =~ /xml/ && editorIsOpen && XMLLSReady"
+        },
+        {
+          "command": "xml.restart.language.server",
```
datho7561, line 719: I still see restart in the command palette when I open a blank window before opening any XML files.

datho7561 (changes requested):

src/client/xmlClient.ts
angelozerr: I tried it but it doesn't work since it is a finction in the map.

README.md
```diff
@@ -20,6 +20,7 @@ This VS Code extension provides support for creating and editing XML documents,
 | enabled by default | requires additional configuration to enable |
 
   * [RelaxNG (experimental) support](https://github.com/redhat-developer/vscode-xml/blob/main/docs/Features/RelaxNGFeatures.md#relaxng-features) (since v0.22.0)
+  * [Surround with Tags,Comments, CDATA](https://github.com/redhat-developer/vscode-xml/blob/main/docs/Refactor.md#refactor) (since v0.23.0)
```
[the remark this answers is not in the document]
angelozerr: fixed

docs/Refactor.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Surround with Tags (Wrap)
+
+This refactor command gives the capability to select an XML content and surround it with a given tag. To execute this command you can:
+
+ * use command palette (`Ctrl+P`) and type `Surround`
+
+![Surround with Tags](images/Refactor/SurroundWithTags.gif)
+
+ * use contextual menu
```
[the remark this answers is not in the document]
angelozerr: I add a comment, please tell me if it is ok

docs/Refactor.md
angelozerr: fixed

docs/Refactor.md
angelozerr: fixed

package-lock.json
```diff
@@ -1285,7 +1285,7 @@
     },
     "buffer-crc32": {
       "version": "0.2.13",
-      "resolved": "https://registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
+      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
```
rgrunber, line 1288: Don't forget to remove `repository.engineering.redhat.com/nexus/repository/` segment from the resolved entry.

rgrunber:
A few considerations :

- Activating functionality through code actions is a pretty common work flow even when selecting text. Would it be possible to create a code action for the "surround with ..." command so that it can be triggered with `<ctrl> + .` as opposed to through the command palette ? Code actions can be commands, so would it be possible to return a command the client will handle, whose arguments are the snippet (if supported).
- Are there cases where we can perform the surround without any selection being made ? If a user makes no selection, but the cursor is within a tag (eg. `<som|eTag>`, would it make sense to have the command surround from the start of `<someTag>` all the way to `</someTag>` ? If the end tag is not present, maybe just surrounding `<someTag>` ?

Let me know your thoughts. If you agree with any, we could always look at them after merging this first iteration.

src/client/xmlClient.ts
datho7561: It seems like it works if you wrap the return object in `(` `)`

package.json
datho7561: I was running an out  of date version of vscode-xml it seems

datho7561:
> Would it be possible to create a code action for the "surround with ..." command so that it can be triggered with `<ctrl> + .` in addition to command palette ?

+1 I think this would be a good idea for a future PR, it would involve figuring out how to pass what range the user has selected.

> Are there cases where we can perform the surround without any selection being made ?

Yes, and I think the current behaviour is pretty helpful for quickly writing XML.

> If a user makes no selection, but the cursor is within a tag (eg. `<som|eTag>`), ...

@otichy suggested "If the selection leads to a malformed xml, that's the user's problem." when there is a selection, and I agree with this. However, if the cursor is on an open/close tag, I agree that it could be neat to wrap the element, especially now that we are contacting the language server anyways. I think this would be a good idea for a future PR.

angelozerr:
> Activating functionality through code actions is a pretty common work flow even when selecting text. 

+1

> Are there cases where we can perform the surround without any selection being made ? If a user makes no selection, but the cursor is within a tag (eg. <som|eTag>), would it make sense to have the command surround from the start of <someTag> all the way to </someTag> ? If the end tag is not present, maybe just surrounding <someTag> ?

I wanted to take care of that, but it seems according to @otichy, Oxygen doesnt take care of that. My suggestion was to adjust selection according to a setting https://github.com/redhat-developer/vscode-xml/pull/813#issuecomment-1339800245

I agree that surrounf can be improved, but please see that in the future PR. I would like to merge this PR ASAP in order @otichy can test it and give us feedback before the release.

package-lock.json
angelozerr: fixed

src/commands/serverCommandConstants.ts
angelozerr: fixed

angelozerr:
> If I understand @rgrunber and as @datho7561 is probably pointing out, these are two different cases. If the user actually makes a selection, it should be respected, even if the resulting xml is not well-formed. However, if the user makes no selection and the cursor is merely somewhere on the opening/closing tag itself (but not between those opening/closing tags), it makes sense to wrap the whole element. That's also how oXygen does it. But frankly, I have never noticed that before I tried it just now, so I wouldn't consider that an important feature.

I created a PR for that https://github.com/eclipse/lemminx/pull/1392, see demo at https://github.com/eclipse/lemminx/pull/1392#issuecomment-1341716540

datho7561 (approved):
Looks good to me. Thanks Angelo! Sorry the review process took so long.

PR #814, opened 2022-11-18
lemminx logfile: fallback to global storage

mkhl:
when opening an xml file outside of a workspace, this extension always create a folder `~/.lemminx` to store the language server log file (even when the `xml.server.workDir` preference is set to some other directory)

since it uses the workspace storage in the context of a workspace, it seems more sensible to fallback to the global storage instead.
(i also consider creating random folders in users' homes bad form but that's just preference)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -13,7 +13,7 @@
 import * as fs from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { ExtensionContext, extensions, languages } from "vscode";
+import { ExtensionContext, Uri, extensions, languages } from "vscode";
 import { Executable, LanguageClient } from 'vscode-languageclient/node';
 import { XMLExtensionApi } from './api/xmlExtensionApi';
 import { getXmlExtensionApiImplementation } from './api/xmlExtensionApiImplementation';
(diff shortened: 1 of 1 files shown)


rgrunber:
@datho7561 , let me know if you see any downsides to this. I think it makes sense to avoid placing things in the user's home folder if possible.

Also `xml.server.workDir` seems like a really weird property name for cached schema storage. I wonder if it makes sense to have that also default to the plugin's global storage location.

datho7561:
I'll take a look at it. The default cache directory for lemminx can't be in the VS Code global workspace (because it might not be running in VS Code), but I think it makes sense to set the cache directory to be in the global workspace when running from VS Code.

datho7561:
@mkhl it looks good, and I think this is the right approach. There are some linting and compile errors. When you get a chance, could you please address them?

mkhl:
> @mkhl it looks good, and I think this is the right approach. There are some linting and compile errors. When you get a chance, could you please address them?

done. i just had to find the "Fix package-lock.json" ci step to get it working 😅

datho7561 (approved):
Looks good. Thanks @mkhl !

PR #824, opened 2022-11-25
Combine 'startup' and 'server.initialized' telemetry events

### LINKED ISSUE #818: Combine "startup" & "server.initialized" telemetry events
If a user has opted-in, we send 2 events on startup of the extension. One purely from the client side, and one from the server side. It would be good to combine these into one. Maybe the "startup" event should wait for the "server.initialized" event, or something to that effect.

Main problem is that, if "server.initialized" fails, there's no server to send the event for failure, so "startup" shouldn't wait. Either some timeout, or a way for the client to check if the server failed.

JessicaJHee:
* in the case where the `server.initialized` telemetry is successfully received by the client, it sends a `startup` telemetry including the properties in the `server.initialized` sent by the server side.
* if `server.initialized` is never received by the client, there is a 15 second time out to catch this, then an empty `startup` telemetry is sent with no properties. 

Fixes #818 

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -1,5 +1,6 @@
 import { getRedHatService, TelemetryService } from "@redhat-developer/vscode-redhat-telemetry";
 import { ExtensionContext } from "vscode";
+import { getXMLConfiguration } from "./settings/settings";
 
 /**
  * Wrap vscode-redhat-telemetry to suit vscode-xml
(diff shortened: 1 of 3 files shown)


src/telemetry.ts
```diff
@@ -19,6 +19,7 @@ export const BINARY_DOWNLOAD_FAILED = "failed";
 export const BINARY_DOWNLOAD_ABORTED = "aborted";
 
 let _telemetryManager: TelemetryService = null;
+let serverInitialized = false;
```
rgrunber, line 22: I would call this `serverInitializedReceived` for now.

rgrunber (commented):
Looks pretty good. Could you also combine the `preferBinary` value that's sent as part of the `Telemetry.SETTINGS_EVT` into the `startup` event ? We only send it after startup also.

rgrunber (approved):
I think it makes sense to merge. Just one small thing to keep in mind for very slow machines :

1. You open some XML file and it activates the extension & language server
2. The extension  begins waiting for the "server.initialized", and it never comes, so it sends a simple "startup".
3. "server.initialized" finally arrives, very slowly, so an additional "startup" is sent with the missing data from the first "startup".

We could check for this at `  if (eventName === STARTUP_EVT) { .. }` but then we'd miss the additional data. So I think in this case, it should be ok to send the additional event. The vast majority of clients will have a language server respond within the timeout.

PR #830, opened 2022-12-06
Add codelens and command to register/unregister catalog

### LINKED ISSUE #572: CodeLens for registering/unregistering catalog
It should be nice to insert in catalog xml file a Codelens `[Register catalog]` to insert it to the list of catalog (see [xml.catalogs](https://github.com/redhat-developer/vscode-xml/blob/master/docs/Preferences.md#catalogs)).

```xml
[Register catalog]
<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
  <uri name="http://docs.oasis-open.org/odata/ns/edmx" uri="./schemas/edmx.xsd" />
  <uri name="http://docs.oasis-open.org/odata/ns/edm" uri="./schemas/edm.xsd" />
</catalog>
```

When catalog.xml is already registered, it should be nice to have a `[Unregister catalog]` codelens.

It should be good to have a vscode command and quickfix for that too.

JessicaJHee:
Fixes #572 
Requires https://github.com/eclipse/lemminx/pull/1390


datho7561:
Something that I ran into is that if the catalog is enabled in the user settings, then the unregister catalog button doesn't work.

JessicaJHee:
> Something that I ran into is that if the catalog is enabled in the user settings, then the unregister catalog button doesn't work.

Thank you! Still working on managing user and workspace settings...

src/commands/registerCommands.ts
```diff
@@ -328,6 +329,60 @@ function registerAssociationCommands(context: ExtensionContext, languageClient:
...
+  editType: ConfigurationItemEditType;
+}
+
+enum ConfigurationItemEditType {
+  Add = 0,
+  Delete = 1
+}
+
+/**
+ * Register commands used for registering/unregistering catalog for command menu and CodeLens
```
angelozerr, line 344: It is not only for catalog, it is a command to update configuration on client side.

src/commands/registerCommands.ts
```diff
@@ -328,6 +329,60 @@ function registerAssociationCommands(context: ExtensionContext, languageClient:
...
+ */
+function registerUpdateConfigurationCommands(context: ExtensionContext) {
+  // For CodeLens
+  context.subscriptions.push(commands.registerCommand(ClientCommandConstants.UPDATE_CONFIGURATION, resolveConfigurationItemEdit));
+}
+
+function resolveConfigurationItemEdit<T>(configurationItemEdit: ConfigurationItemEdit) {
+  const document = (window.activeTextEditor && window.activeTextEditor.document && window.activeTextEditor.document.languageId === 'xml') ? window.activeTextEditor.document : undefined;
+  const currentWorkspaceUri = getWorkspaceUri(document).toString();
+  const catalogFilePath = getDirectoryPath(configurationItemEdit.value).includes(currentWorkspaceUri) ? getRelativePath(currentWorkspaceUri, configurationItemEdit.value) : configurationItemEdit.value;
```
angelozerr, line 357: It is not generic because it is linked to catalog. I wonder if this resolution could be done on server side. Or if it is not possible, I think configurationItemEdit should store a kind data to say that it is a file. when kind = file we can resolve the path like you did.

angelozerr:
@JessicaJHee thanks so much to manage catalog with generic update preferences but keep inmind that your code should not be linked to catalog. The idea is that your generic update preferences on client side will be used in the future for another feature.

Here some ideas:

 * when XML is bound to a DTD which have some entity which should be resolved and the setting resolveExternalEntities is set to false, we could provide a CodeAction which say "do you want to enable resolveExternalEntities" ?. Click on this code action will use your generic update configuration command.
 * when there is a proxy problem, we could provide a codeaction which says, do you want to update your proxy

angelozerr:
Could create the same file than https://github.com/redhat-developer/vscode-microprofile/blob/master/src/lsp-commands.ts in vscode-xml and adapt it withyour new feature.

The main idea is that you can copy/paste this ts file and use it in any project.

angelozerr (approved):

angelozerr:
Thanks @JessicaJHee !

PR #837, opened 2022-12-13
Pressing "escape" during binding wizard doesn't stop the binding process

### LINKED ISSUE #832: Pressing "escape" during binding wizard doesn't stop the binding process
eg. press "Escape" when prompted for the binding type when trying to bind a file to a schema. Instead of closing the wizard and ending the binding process, VS Code freezes and the file picker appears.

angelozerr:
Pressing "escape" during binding wizard doesn't stop the binding process

Fixes #832


datho7561 (approved):
Looks good and works well. Thanks, Angelo!

PR #854, opened 2023-01-08
Add snippets for XSLT and XSD

EmilySeville7cfg:


angelozerr:
Thanks for your contribution but snippets are managed on xml language server side to filmer with some condition written in java if snippet must be displayed or not.

With your pr, it will show xsd snippets in an xml file which is wrong.

If yiu need some assistance to manage snippet on lemminx dont hesitate to ask us.

angelozerr:
The idea is to register snippet on server side like [lemminx](https://github.com/eclipse/lemminx)/[org.eclipse.lemminx](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx)/[src](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src)/[main](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/main)/[resources](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/main/resources)/[org](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/main/resources/org)/[eclipse](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/main/resources/org/eclipse)/[lemminx](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx)/[services](https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services)/snippets

The main problem with vscode snippet is that you can register it for a given language but you cannot show the snippet only if for instance the snippet is trigerred inside à given element tag name.

More it works only for vscode. Another lsp client cannot benefit from snippets registered in vscode.

It is reason why we manage snippet on server side.

snippets/xsd-snippets.json
```diff
@@ -0,0 +1,16 @@
+{
+  "el": {
+    "description": "[el]ement",
+    "prefix": "el",
+    "body": [
+      "<xs:element name=\"${1:string}\" type=\"${2|xs:string,xs:decimal,xs:integer,xs:boolean,xs:date,xs:time|}\" default=\"${1:string}\"/>"
```
angelozerr, line 6: XSD completion which is based on XSD xml.xsd provides already element and attribute for completion, what is the benefit of those snippets?

snippets/xslt-snippets.json
```diff
@@ -0,0 +1,41 @@
+{
```
angelozerr, line 1: XSLT completion support based on XSD schema for XSL is implemented and provides those kind of completion items, what is the benefit of those snippet?

snippets/xsd-snippets.json
EmilySeville7cfg: Does completion provides all attributes mentioned in snipped as so quickly as snippet does?

snippets/xslt-snippets.json
EmilySeville7cfg: To reduce keystrokes required to create elements and attributes with prefilled attributes.

snippets/xsd-snippets.json
angelozerr: > Does completion provides all attributes mentioned in snipped as so quickly as snippet does?

It provides only required attributes. Having too many completion items is not helpfull. So we should implement a java filter to show xs:element only if it is allowed. 

> BTW, why there is no XSD choice when changing language in VS Code?

It is a good sample why it is not interesting to support snippets with vscode, xsd:choice is only available of xs:complexType, so it should appear only when completion is triggered inside xs:complexType. As XSD completion is based onXSD schema, it show xsd:choice only when it is allowed.

angelozerr: > I wanna check how completion works.

Create a new file with `test.xsd`. XSD uses XML as language (and not XSD).

EmilySeville7cfg: > It provides only required attributes. Having too many completion items is not helpfull. So we should implement a java filter to show xs:element only if it is allowed.

Then I guess my request is invalid.

angelozerr: Once you will creat ethis file,you can opencompletion and you will see snippets:

image

You will notice that there are few snippets and not the all snippets of lemminx. We filter the snippet available only when a file is empty and have XSD extension.

EmilySeville7cfg: Sounds great. :)

angelozerr: If you open completion, you will see that xsd:choice is not avialable inside xsd:element:

image but it is available inside complexType:

image

snippets/xslt-snippets.json
pgfearo: @EmilySeville7cfg Perhaps you're not ware that there's an [XSLT/XPath language extension](https://marketplace.visualstudio.com/items?itemName=deltaxml.xslt-xpath) for VS Code? 

This extension has context-aware auto-completion snippets for all XSLT 3.0 (and the proposed XSLT 4.0 instructions) - the snippets include attributes and child-elements where required. As well as XML validation, XPath expressions, variables, functions etc. are also validated.

PR #857, opened 2023-01-13
React to vsce renaming to @vscode/vsce.

rgrunber:
- Use --force to override any older vsce provided by NodeJS tooling


datho7561 (approved):
Looks good. Thanks for making this change, Roland!

PR #862, opened 2023-01-23
Support identation for XML contributed languages

angelozerr:
Support identation for XML contributed languages

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -36,8 +36,12 @@ export async function activate(context: ExtensionContext): Promise<XMLExtensionA
 
   registerClientOnlyCommands(context);
 
-  languages.setLanguageConfiguration('xml', getIndentationRules());
-  languages.setLanguageConfiguration('xsl', getIndentationRules());
+  // Update indentation rules for all language which are XML.
+  XML_SUPPORTED_LANGUAGE_IDS.forEach(l => {
+    const languageId = <string> l;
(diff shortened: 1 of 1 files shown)


datho7561:
There is a bug with linked editing that makes proper tag indentation very annoying. For instance, if you have an `.svg` document:

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.w3.org/2000/svg https://www.w3.org/TR/2002/WD-SVG11-20020108/SVG.xsd">
  |
</svg>
```

If you add a tag `<aaa>` at the `|`, hit enter, then type in `</aaa>`, it will remove content from the opening tag when you type the last `>`.

This behaviour doesn't happen when you have tag auto closing, since the indentation isn't triggered if you have auto tag closing (eg. in xml and xsd files), but `.svg` files don't have auto indentation configured by default. However, if you disable auto tag closing, you can see the behaviour in XML and XSD files as well.

datho7561:
broken-indentation

angelozerr:
@datho7561 I created a PR to try to fix issue with linked editing range at https://github.com/eclipse/lemminx/pull/1457

But the bug that you found is not related to this vscode-xml PR indentation.

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #866, opened 2023-01-27
Publish a generic vsix for all other platforms.

### LINKED ISSUE #865: Should publish a generic vsix
Only a subset of platforms are covered here, instead of publishing a generic vsix. 

https://github.com/redhat-developer/vscode-xml/blob/a17945188ade7f39a3d8ee0fbc75443db5ac3200/Jenkinsfile#L127

Similar to https://github.com/redhat-developer/vscode-java/pull/2880

rgrunber:
- Instead of publishing a generic vsix for marketplace platforms that we cannot support, simply continue to publish a generic vsix
- Publish platform-specific extensions, and pre-releases on Open VSX
- Fixes #865

Note that the reason the `--packagePath` publishing calls don't have any pre-release flags, is because the flag is embedded in the vsix itself at build-time. Only the generic vsix publishing needs the pre-release flag at publish time.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -106,9 +106,9 @@ node('rhel8'){
 node('rhel8'){
 	sh 'npm install -g --force "@vscode/vsce"'
 	sh 'npm install -g "ovsx"'
-	if(publishToMarketPlace.equals('true') || publishPreRelease.equals('true')){
+	if(publishToMarketPlace.equals('true') || publishToOVSX.equals('true') || publishPreRelease.equals('true')){
 
-		if (publishToMarketPlace.equals('true')) {
+		if (publishToMarketPlace.equals('true') || publishToOVSX.equals('true')) {
 			timeout(time:2, unit:'DAYS') {
 				input message:'Approve deployment?', submitter: 'fbricon,rgrunber,azerr,davthomp'
 			}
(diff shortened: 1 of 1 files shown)


fbricon (approved):

Jenkinsfile
```diff
@@ -117,22 +117,29 @@ node('rhel8'){
...
 		unstash 'vsix'
 		def vsix = findFiles(glob: '**.vsix')
+		def platformVsixes = findFiles(glob: '**.vsix', excludes: vsix[0].path)
+
 		// VS Code Marketplace
-		withCredentials([[$class: 'StringBinding', credentialsId: 'vscode_java_marketplace', variable: 'TOKEN']]) {
-			def platformVsixes = findFiles(glob: '**.vsix', excludes: vsix[0].path)
-			for(platformVsix in platformVsixes){
-				sh 'vsce publish -p ${TOKEN}' + " --packagePath ${platformVsix.path}"
+		if (publishToMarketPlace.equals('true')) {
```
datho7561, line 123: By wrapping this in an if, the prereleases won't publish, since it seems like using `publishToMarketPlace` signals that this is a release build. If instead `publishToMarketPlace` is just meant to signal where to publish, then the popup to click "approve" will appear during prereleases (line 111), blocking the build.

Jenkinsfile
```diff
@@ -117,22 +117,29 @@ node('rhel8'){
...
+				sh 'vsce publish -p ${TOKEN}' + " ${env.publishPreReleaseFlag}"
 			}
-			// Cannot combine packagePath & target, so re-generate (generic) package and publish
-			sh 'vsce publish -p ${TOKEN} --target win32-ia32 win32-arm64 linux-arm64 linux-armhf alpine-x64 alpine-arm64' + " ${env.publishPreReleaseFlag}"
 		}
 
-		if (publishToMarketPlace.equals('true')) {
-			// Open-VSX Marketplace does not support pre-release
+		// Open-VSX Marketplace
+		if (publishToOVSX.equals('true')) {
```
datho7561, line 134: Same thing here

datho7561 (changes requested):

Jenkinsfile
rgrunber: Right. When only `publishPreRelease` is set, it should still publish pre-release to all marketplaces.

rgrunber: I just added `|| publishPreRelease.equals('true')` to the ifs. That should allow pre-releases to always publish but just selecting that, regardless of marketplace, while allowing us to do proper releases to individual marketplaces. Let me know if I got something wrong.

datho7561 (approved):
Looks good!

PR #871, opened 2023-01-31
Upversion to 0.25.0

rgrunber:
Let's go to 0.25.0 for now so that we continue using that in pre-releases (as opposed to something containing 1.0.0). Once we get approval to go 1.0.0, we can change the version just before releasing.


datho7561 (approved):
Looks good! Thanks, Roland.

PR #894, opened 2023-04-17
Update VS Code engine to `^1.67.0`

### LINKED ISSUE #872: Upgrade engine to "^1.67.0"
In the package.json, we say that we require VS Code "^1.63.0", but since we upgraded to a newer language client, we actually rely on VS Code "^1.67.0". We should upgrade the listed engine version so that older versions of VS Code are prevented from installing the extension.

datho7561:
Fixes #872

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     }
   },
   "engines": {
-    "vscode": "^1.63.0"
+    "vscode": "^1.67.0"
   },
   "activationEvents": [
     "onLanguage:xml",


JessicaJHee (commented):
LGTM! This is from https://github.com/microsoft/vscode-languageserver-node/blob/ffaf05ddd1fa9ca5f258cd0b08cc6cd1794c5b57/client/src/node/main.ts#L26 since 8.0.0?

JessicaJHee (approved):

datho7561:
Cool! Thanks for the review, Jessica!

PR #912, opened 2023-06-30
Changelog for 0.26.0

rgrunber:
- Upversion to 0.26.0
- Use LemMinX 0.26.1

Do we take advantage of the 2 enhancements coming from the LemMinX side in vscode-xml ? We don't have any code actions that aren't associated with a diagnostic ? Our rename functionality doesn't currently have any cases for functioning across documents ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## [0.26.0](https://github.com/redhat-developer/vscode-xml/milestone/34?closed=1) (July 10, 2023)
+
+### Bug Fixes
+
+ * `xml.foldings.includeClosingTagInFold` has no effect in binary mode. See [#1523](https://github.com/eclipse/lemminx/issues/1523).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Change Log
 
+## [0.26.0](https://github.com/redhat-developer/vscode-xml/milestone/34?closed=1) (July 3, 2023)
+
+### Enhancements
+
+ * Add support for code actions that are not associated with any diagnostics (quick assist). See [#1516](https://github.com/eclipse/lemminx/issues/1516).
+ * Make the rename participants conform to the LSP, permitting to rename across different documents. See [#1521](https://github.com/eclipse/lemminx/issues/1521).
```
datho7561, line 8: I think you are right, I don't think there is any user-visible improvements from these changes in vscode-xml. I think we should remove these from the changelog.

datho7561 (changes requested):

CHANGELOG.md
rgrunber: I've removed them.

PR #933, opened 2023-09-18
Update GH Actions release workflow

rgrunber:
The main purpose of this change is to add the ability to publish a pre-release when there are changes to either lemminx or vscode-xml. This is done by having a daily scheduled workflow that has a job which checks out the 2 repositories, and determines if the recent commit (from both repos) have already been encountered in a previous build. This check is skipped if the build was manually triggered (assume the maintainer knows what they're doing).

- https://github.com/DeLaGuardo/setup-graalvm/ has been archived since April 2023, so we should use https://github.com/graalvm/setup-graalvm
- No need for workflow to release binary artifacts of lemminx in the native-image job, because we already do it in vscode-xml releases, (eg. https://github.com/redhat-developer/vscode-xml/releases/tag/0.26.1 .

- [x] Verify that the new binaries (resulting from the new setup-graalvm with version 22.3.3) work as expected.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: release
 
 on:
+    schedule:
+      - cron: '0 8 * * *'
     workflow_dispatch:
       inputs:
         publishPreRelease:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -28,7 +30,38 @@ on:
             - 'false'
           default: 'false'
 jobs:
+  should-build-change:
+    runs-on: ubuntu-latest
+    outputs:
+        repo-cache-hit: ${{ steps.should-build-change.outputs.REPO_CACHE_HIT }}
```
rgrunber, line 36: So `should-build-change` (job) has an "output" called `repo-cache-hit` which is set to the value of the `REPO_CACHE_HIT` (env) set in the `should-build-change` (step). It should be possible to get rid of one "chain" here.

What if the output for `should-build-change` (job) was just set to `steps.cache-last-commit.outputs.cache-hit == 'true'` ?

rgrunber:
Will be merging this later today. Based on my testing this should work as expected.

PR #948, opened 2023-11-16
Update to Node 18

### LINKED ISSUE #928: Node 14 is EOL
Node 16 will also be EOL (no maintenance) on September 11 2023. As a result, we should upgrade to Node 18.

A part of this migration is updating the version of npm used and updating `package-lock.json`. npm v7, which ships with Node 16, changed the format of this file. This might cause some complications with `npm-force-resolutions`, which is a library we use to for all instances of a library to a specified version in the lockfile.

rgrunber:
- Fixes #928 
- Updated typescript-eslint-plugin to latest version for proper Typescript 4.9 support
- I didn't update Jenkinsfile as that's basically obsolete


datho7561 (approved):
Looks good! Thanks, Roland!

Issue #898, opened 2023-04-23
Mac ARM / AS


ST33LDI9ITAL:
Mac users shouldn't have to use Rosetta for a vscode plugin. That's ridiculous.

thtliife:
@ST33LDI9ITAL... I can use it without rosetta on an m2 MacBook.
Steps to get it going:

1. Install [homebrew](https://docs.brew.sh/Installation)

2. Install openjdk:

```bash
brew install openjdk

# get the JAVA_HOME path in your clipboard
echo $(brew --prefix openjdk)/libexec/openjdk.jdk/Contents/Home | pbcopy
```

3. Add java home path to vscode settings for Xml Language server
* Open vscode settings: <kbd>⌘</kbd> + <kbd>.</kbd>
* In search settings bar type: `XML > Java: Home`
* Click link for Edit in settings.json
* Paste the previously copied path for `JAVA_HOME` in between the empty quotes provided in the automatically inserted `"xml.java.home": ""` property.

4. Restart VSCode

fbricon:
@datho7561 @rgrunber  Github actions now provide M1 runners https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/, we should be able to generate the LemMinX binary for ARM-based macs.

rgrunber:
Could someone verify that https://github.com/rgrunber/vscode-xml/releases/download/0.26.1/vscode-xml-darwin-arm64-0.26.1-693.vsix works as expected without Rosetta ?

PR #980, opened 2024-03-21
Remove stale comment about Apple ARM

fbricon:


rgrunber (approved):

rgrunber:
Yup, https://github.com/redhat-developer/vscode-xml/issues/898 is fixed.

fbricon:
@rgrunber https://github.com/redhat-developer/vscode-xml/pull/980/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R61 will soon need to go too right?

rgrunber:
Was thinking that.. but no. On all Linux distros (except Alpine), we still require users to have libc because we're keeping it dynamically linked. It's just on Alpine that we won't. Is it worth adding a "except on Alpine x64" to that line ?

fbricon:
yeah probably, for the sake of proper doco, as there's no other mention of alpine

PR #985, opened 2024-04-10
fix the error that "/" in attribute value make indentation problem

wangxi761:
https://github.com/redhat-developer/vscode-xml/blob/d4db83adf421ccede3a4264439c1264f69e47b0d/src/client/indentation.ts#L14

The current regex pattern used for matching non-self-closing XML tags fails to match tags that include attributes. Specifically, the pattern struggles with tags formatted like `<a type="text/json">`, which are not self-closing but contain attributes within them. The issue stems from the regex not adequately accounting for the presence of `/` characters within attribute values, which leads to incorrect or failed matches.

### DIFF
--- a/src/client/indentation.ts
+++ b/src/client/indentation.ts
@@ -11,12 +11,12 @@ export function getIndentationRules(): LanguageConfiguration {
     },
     onEnterRules: [
       {
-        beforeText: new RegExp(`<([_:\\w][_:\\w-.\\d]*)([^/>]*(?!/)>)[^<]*$`, 'i'),
+        beforeText: new RegExp(`<([_:\\w][_:\\w-.\\d]*)(([^/>]|/[^>]*["'])*(?!/)>)[^<]*$`, 'i'),
         afterText: /^<\/([_:\w][_:\w-.\d]*)\s*>/i,
         action: { indentAction: IndentAction.IndentOutdent }
       },
       {
-        beforeText: new RegExp(`<(\\w[\\w\\d]*)([^/>]*(?!/)>)[^<]*$`, 'i'),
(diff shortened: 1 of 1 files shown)


angelozerr:
If I remember I did that by adjusting vscode html beforetext https://github.com/microsoft/vscode/blob/6a06de0033d052521f03079aabb356ca7916145c/extensions/html/language-configuration.json#L37

But I think more and more that we should defines that in the language configuration like vscode html does.

@datho7561 what do you think about this PR? About my suggestion?

datho7561:
I think that we should use what is configured for html, but with the first negative lookahead removed so we don't exclude those tags from indentation. eg. (note that `"` is being escaped in the following regex):

```regex
<([_:\\w][_:\\w-.\\d]*)(?:(?:[^'\"/>]|\"[^\"]*\"|'[^']*')*?(?!\\/)>)[^<]*$
```

We might also have to adjust this so it doesn't indent for process instructions eg. `<?any content can be here?>` since they don't need to be closed

wangxi761:
As an aside, do we need to add non-capturing symbol `?:` to every capturing group to increase the speed of the regular expression

datho7561:
MDN says that using non-capturing groups will speed up the regex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Non-capturing_group

The only reason I see not to use non-capturing groups is that they make it harder to read the regex. Then again, it's not much more difficult to read, and if you use a dedicated tool like regex101.com to test the expressions it has tools to help pick apart what's actually helping.

wangxi761:
I modified the regular expression  and add non-capturing symbols for capturing group

Issue #986, opened 2024-04-15
Alpine support


cirolosapio:
as [eclipse/lemminx](https://github.com/eclipse/lemminx?tab=readme-ov-file#generating-a-native-binary) readme says: 

> On Linux, compile with ./mvnw clean package -Dnative -DskipTests -Dgraalvm.static=--static in order to support distributions that don't use glibc, such as Alpine Linux

is it possible to add an alpine build with the `-Dgraalvm.static=--static` ?

cirolosapio:
is there a released version of the vscode extension that use the compiled lemminx with that property?

fbricon:
No there is not. We can't release 2 linux versions to the marketplace. 
Ideally, the binary would be generated and stored as a release artifact in this repo, but our team is stretched too thin at the moment so we can't commit to address this issue anytime soon.

rgrunber:
At first we had the static flag to support https://github.com/redhat-developer/vscode-xml/issues/457 . Then we realized in https://github.com/redhat-developer/vscode-xml/issues/597#issuecomment-930343104 that the flag broke for all other linux distros.

However, I just had a look at my old comment at https://github.com/redhat-developer/vscode-xml/issues/597#issuecomment-952988946 . VS Code does support `alpine-x64`. If all that's needed is a a static glibc, and if we can just use a GH Actions Ubuntu runner that simply has the static flag enabled, I think it could work.

But yeah, it adds yet another part to the build+release process.

cirolosapio:
something like this #989 ?

fbricon:
Alpine extension is now available in the VS Code Marketplace

<img width="1240" alt="Screenshot 2024-04-24 at 12 48 59" src="https://github.com/redhat-developer/vscode-xml/assets/148698/4d108bf3-ff71-4571-9d3b-4acc30c5e4e4">

As well as https://github.com/redhat-developer/vscode-xml/releases/tag/latest

PR #989, opened 2024-04-20
Alpine support

cirolosapio:
- Fixes https://github.com/redhat-developer/vscode-xml/issues/986

### DIFF
--- a/.github/workflows/native-image.yaml
+++ b/.github/workflows/native-image.yaml
@@ -34,7 +34,7 @@
               label: 'osx-aarch_64'
             - os: ubuntu-20.04
               label: 'linux'
-              props: -Dgraalvm.static=-H:+StaticExecutableWithDynamicLibC
+              prop: -Dgraalvm.static=-H:+StaticExecutableWithDynamicLibC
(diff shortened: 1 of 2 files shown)


rgrunber:
Yup! this is almost exactly what I had locally. Still haven't fully tested though. I think https://github.com/redhat-developer/vscode-xml/pull/989/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L119 also needs to become `lemminx-{linux,alpine}` because the generated binary take the name of the `matrix.label`.

cirolosapio:
reopened #991

rgrunber:
I noticed the same issue.. for some reason when I tried doing a regular release, there was an issue uploading the artifact I think your approach of making the label, the "primary key" is the right one. I guess I must have only tested when alpine was on ubuntu-20.04 and linux was on ubuntu-22.04, and never tried having both on the same value.

cirolosapio:
with this approach it compiles the right binary https://github.com/cirolosapio/vscode-xml/actions/runs/8808258955/job/24177003106

PR #999, opened 2024-05-28
Use the extension version in download URL for binary.

rgrunber:
- The language server version was used as the version for the download URL of the binary, but since the binaries are hosted under the vscode-xml release tab, it should match the vscode-xml release
- Remove some unnecessary steps from the release job

I also slightly refactored the ordering. The preparing of the pre-release affects the final version number, so extracting the version number must happen after. Any workflow that users the versions can occur immediately after.

Also set xmlServer/version back to 0.28.0 as we may end up doing a bug fix release for this, using the same language server.


datho7561 (approved):
Seems good to me

PR #1016, opened 2024-07-08
Fix alpine lemminx server

### LINKED ISSUE #1015: lemminx-alpine server not found
the alpine version is present
image

but the extension is looking for the linux one
```
[Error - 11:49:40 AM] XML Support client: couldn't create connection to server.
Launching server using command /home/vscode/.vscode-server-insiders/extensions/redhat.vscode-xml-0.27.2024053103-alpine-x64/server/lemminx-linux failed. Error: spawn /home/vscode/.vscode-server-insiders/extensions/redhat.vscode-xml-0.27.2024053103-alpine-x64/server/lemminx-linux ENOENT
```

cirolosapio:
fixes #1015

### DIFF
--- a/src/server/binary/binaryServerStarter.ts
+++ b/src/server/binary/binaryServerStarter.ts
@@ -219,7 +219,7 @@ function getPlatform (): ReturnType<typeof os.platform> | 'alpine' {
     const lines = osRelease.split('\n')
     for (const line of lines) {
       const [key, value] = line.split('=')
-      if (key === 'ID') return value as 'alpine'
+      if (key === 'ID' && value === 'alpine') return value
     }
   }
   return os.platform()
--- a/src/server/binary/binaryServerStarter.ts
+++ b/src/server/binary/binaryServerStarter.ts
@@ -214,14 +214,7 @@ function getBinaryEnvironment(): any {
  * @returns The platform of the current operating system.
  */
 function getPlatform (): ReturnType<typeof os.platform> | 'alpine' {
-  if (fs.existsSync('/etc/os-release')) {
-    const osRelease = fs.readFileSync('/etc/os-release', 'utf8')
-    const lines = osRelease.split('\n')
-    for (const line of lines) {
-      const [key, value] = line.split('=')
-      if (key === 'ID' && value === 'alpine') return value
-    }
-  }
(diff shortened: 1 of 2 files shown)


src/server/binary/binaryServerStarter.ts
```diff
@@ -206,13 +206,32 @@ function getBinaryEnvironment(): any {
...
 
+/**
+ * Returns the platform of the current operating system.
+ * If the OS is Alpine Linux, it returns 'alpine'.
+ * Otherwise, it returns the platform string provided by the 'os' module.
+ *
+ * @returns The platform of the current operating system.
+ */
+function getPlatform (): ReturnType<typeof os.platform> | 'alpine' {
+  if (fs.existsSync('/etc/os-release')) {
```
rgrunber, line 217: Would it be sufficient to just use the presence of `/etc/alpine-release` ? to return `alpine` ? I'd like to keep this as minimal as possible without any reading.

src/server/binary/binaryServerStarter.ts
```diff
@@ -206,13 +206,32 @@ function getBinaryEnvironment(): any {
...
 }
 
+/**
+ * Returns the platform of the current operating system.
+ * If the OS is Alpine Linux, it returns 'alpine'.
+ * Otherwise, it returns the platform string provided by the 'os' module.
+ *
+ * @returns The platform of the current operating system.
+ */
+function getPlatform (): ReturnType<typeof os.platform> | 'alpine' {
```
rgrunber, line 216: Could this be `: NodeJS.Platform | 'alpine'` or do you feel it's more correct to reference the method itself since that's what is really being called ?

src/server/binary/binaryServerStarter.ts
cirolosapio: thanks, I didn't know that file exists. updated

rgrunber (approved):

cirolosapio:
hi @rgrunber is there an ETA when this bugfix will be released ?

cirolosapio:
i use devcontainer and the extensions list are defined in the `devcontainer.json` file and it is not possible to specify the pre-release version, so it's never installed

datho7561:
Supposedly: https://github.com/microsoft/vscode/issues/149383#issuecomment-1132192731

PR #1034, opened 2024-11-05
Remove 'file' sheme from documentSelector

### LINKED ISSUE #1033: No xml support when opening files on shared folders
Hellò, 
Version: 1.95.1 (user setup)
Commit: 65edc4939843c90c34d61f4ce11704f09d3e5cb6
Date: 2024-10-31T05:14:54.222Z
Electron: 32.2.1
ElectronBuildId: 10427718
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.19045

XML 0.27.1 installed

When I open an xml file and go for formatting i receive the message "no xml formatter found" and i am ask to install one that is already installed.

In activating the extension in the output i read:
[info] ExtensionService#_doActivateExtension redhat.vscode-xml, startup: false, activationEvent: 'onLanguage:xml'
[warning] [redhat.vscode-xml] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '[xml]', provide the URI of a resource or 'null' for any resource.

angelozerr:
Remove 'file' sheme from documentSelector

Fixes #1033

### DIFF
--- a/src/client/languageParticipants.ts
+++ b/src/client/languageParticipants.ts
@@ -59,18 +59,7 @@ export function getLanguageParticipants(): LanguageParticipants {
 
   return {
     onDidChange: onDidChangeEmmiter.event,
-    get documentSelector() {
-      return Array.from(languages) //
-        .flatMap(language => {
-          return [{
-            language,
-            scheme: 'file',
-          }, {
-            language,
-            scheme: 'untitled',
-          }] as DocumentFilter[];
-        });
-    },
+    get documentSelector() { return Array.from(languages); },
     hasLanguage(languageId: string) { return languages.has(languageId); },
     dispose: () => changeListener.dispose()
   };


src/client/languageParticipants.ts
```diff
@@ -63,8 +63,7 @@ export function getLanguageParticipants(): LanguageParticipants {
       return Array.from(languages) //
         .flatMap(language => {
           return [{
-            language,
-            scheme: 'file',
+            language
           }, {
             language,
             scheme: 'untitled',
```
datho7561, line 69: There are two entries in this array being returned. If we don't need to specify the scheme, then we can return just the one entry. In that case, we can replace the `flatMap` with a `map`.

datho7561 (changes requested):

src/client/languageParticipants.ts
angelozerr: I did the change which is the same than vscode html.

datho7561 (approved):
Looks good and works well. Thanks, Angelo!

angelozerr:
Thanks @datho7561 for your great suggestion.

Issue #154, opened 2019-06-07
Plugin is not recognized as a XML formatter


peterkeller:
Even if the plugin is installed, it isn't recognized as a suitable XML formatter when trying to format a XML document, but the message "There is no formatter for 'xml'-files installed" is shown instead.

<img width="1068" alt="Screenshot 2019-06-07 at 09 55 52" src="https://user-images.githubusercontent.com/1429336/59089523-a6a4b080-890a-11e9-8f95-de24a4c4796d.png">

NikolasKomonen:
Could you try disabling all extensions, then enabling only the XML one.

Also just to verify, do you have a JDK installed an setup? Are features like autocompletion or linting working?

Another thing to check is the preference "xml.format.enabled" and make sure that is 'true'

raduspantu:
hello @NikolasKomonen , 

this is the same for me.
I already disabled all my extensions and this is still not working.
![image](https://user-images.githubusercontent.com/24391607/59339958-6af05900-8d05-11e9-875e-e7cbb3e78a1d.png)

Also my JAVA_HOME directory is set 
![image](https://user-images.githubusercontent.com/24391607/59341928-13ec8300-8d09-11e9-9926-dad94572cc08.png)

xorye:
Thank you for reporting this issue.

Could you please go to the command palette (Ctrl+Shift+P), run "Show Running Extensions" and verify that the extension is running?

It should look like this:
![image](https://user-images.githubusercontent.com/20326645/59459265-5c2fa200-8dea-11e9-9248-c518b76854ce.png)


Are features like auto completion and linting working at all?

Are you attempting to format by clicking "Format Document" in the right-click context menu, or are you using Ctrl+Shift+I ? Does the "Format Document" option exist?

peterkeller:
After disabling all extensions and enable this plugin again, everything works as expected. Now, I am enabling the other disabled extensions again. So far, there are no problems.

Thanks.

raduspantu:
@xorye , 

After leaving it alone and focusing on other tasks, I see that now is working, also with other extensions.
What I haven't done until now, is saving the file as an XML, only formatted as an XML; should this be an issue?

Also, Ctrl+Shift+I enables a multi-line for me, but that one I think is a thing from my settings.
It's working now, adding the mention that the file must be saved as an XML before it could recognize the errors, formatting and so on.

Maybe there is something that I am doing wrong...?

xorye:
[...]

PR #1046, opened 2024-12-11
Fix crash of extension when proxy settings address has wrong format

angelozerr:
This PR should fix the following problem:

```
2024-11-22 14:07:45.651 [error] Activating extension redhat.vscode-xml failed due to an error:
2024-11-22 14:07:45.651 [error] TypeError: Cannot read properties of null (reading '1')
	at t.getProxySettings (/Users/userid/.vscode/extensions/redhat.vscode-xml-0.27.2024111708-darwin-arm64/dist/extension.js:2:354116)
```

reported at https://github.com/redhat-developer/vscode-xml/issues/154#issuecomment-2493802486

I think to reproduce the problem, proxy address have a wrong format (don't start with `https://`)

A simple test that I did in debug is:

```typescript
const HOST_AND_PORT_EXTRACTOR = /https?:\/\/([^:/]+)(?::([0-9]+))?/;
HOST_AND_PORT_EXTRACTOR .exec('foo") // -> returns undefined
```

### DIFF
--- a/src/settings/proxySettings.ts
+++ b/src/settings/proxySettings.ts
@@ -75,7 +75,7 @@ export function getProxySettings(): ProxySettings {
     return null;
   }
   const regexResult = HOST_AND_PORT_EXTRACTOR.exec(proxyAddress);
-  if (!regexResult[1]) {
+  if (!regexResult || !regexResult[1]) {
     return null;
   }
   const host: string = regexResult[1];
@@ -152,7 +152,7 @@ export function jvmArgsContainsProxySettings(jvmArgs: string): boolean {
   );
 }
 
-const HOST_AND_PORT_EXTRACTOR = /https?:\/\/([^:/]+)(?::([0-9]+))?/;
+const HOST_AND_PORT_EXTRACTOR = /(?:https?:\/\/)?([^:/]+)(?::([0-9]+))?/;
 
 const JVM_PROXY_HOST = 'http.proxyHost';
 const JVM_PROXY_PORT = 'http.proxyPort';


datho7561 (changes requested):
This change looks pretty good so far.

I think, if we want to support something like `192.168.1.23:9090` as the proxy, we can use this regex:

`(?:https?:\/\/)?([^:/]+)(?::([0-9]+))?`

which just wraps the `https://` part in a non-capturing group and makes it optional.

From what I can tell, we don't use the http:// or https:// part of the url anyways, so it shouldn't matter if it's left out.

angelozerr:
I have tried the const HOST_AND_PORT_EXTRACTOR = (?:https?:\/\/)?([^:/]+)(?::([0-9]+))?;

but it seems it is not a valid regexp

datho7561:
ahh I think you need to surround it in `//`:


```typescript
const HOST_AND_PORT_EXTRACTOR = /(?:https?://)?([^:/]+)(?::([0-9]+))?/;
```

datho7561:
```diff
diff --git a/src/settings/proxySettings.ts b/src/settings/proxySettings.ts
index 3cbcf38..6ee7d0c 100644
--- a/src/settings/proxySettings.ts
+++ b/src/settings/proxySettings.ts
@@ -152,7 +152,7 @@ export function jvmArgsContainsProxySettings(jvmArgs: string): boolean {
   );
 }
 
-const HOST_AND_PORT_EXTRACTOR = /https?:\/\/([^:/]+)(?::([0-9]+))?/;
+const HOST_AND_PORT_EXTRACTOR = /(?:https?:\/\/)?([^:/]+)(?::([0-9]+))?/;
 
 const JVM_PROXY_HOST = 'http.proxyHost';
 const JVM_PROXY_PORT = 'http.proxyPort';
```

angelozerr:
Thanks @datho7561 !

It seems it is working pretty well.

Please merge the PR if you think it is good.

datho7561 (approved):

datho7561:
Thanks for fixing this, Angelo!

PR #1048, opened 2025-01-09
Require Java 11

datho7561:
lemminx is moving to Java 11, so this PR ensures that a Java 11 JRE/JDK is used to run lemminx (or uses the binary if Java 11 is not installed).

This PR also updates the documentation to reflect the new Java 11 requirement.

It also updates the documentation related to instructions on setting up GraalVM for locally building the binary.


rgrunber (commented):
Looks fine overall. Just missing an update at https://github.com/redhat-developer/vscode-xml/blob/f66692877045eb72460987523d6b03686a4af9fb/package.json#L163

rgrunber (approved):
Seems fine for now.

I did run into some odd workflows where the binary gets used automatically when I tried supplying a < 11 Java version. Then when the binary failed (wasn't trusted and I didn't opt to trust it), it complained that the binary server failed and that there is no Java on the system. That last message is technically wrong because there was Java, just not a valid version :stuck_out_tongue_winking_eye: .

datho7561:
https://github.com/redhat-developer/vscode-xml/issues/1049

PR #1050, opened 2025-01-10
If the user specifies an unusable Java, let them know that the binary will be used

datho7561:
Provides a popup that explains what happened and has buttons to:
- Set "prefer binary" to true so they don't see the popup again
- Provide the download link for Java so they can update Java if necessary

This popup only appears when the user has configured `"xml.java.home"` or `"java.home"` (either to an outdated or non-existant JDK/JRE) and this popup doesn't block the language server startup (i.e. we go ahead and launch the binary server before the user clicks anything on the popup).

### DIFF
--- a/src/server/requirements.ts
+++ b/src/server/requirements.ts
@@ -11,6 +11,8 @@ import * as expandHomeDir from 'expand-home-dir';
 const isWindows = process.platform.indexOf('win') === 0;
 const JAVA_FILENAME = 'java' + (isWindows ? '.exe' : '');
 
+export const NO_JAVA_FOUND = "Java runtime could not be located.";
+
 export interface RequirementsData {
   java_home: string;
   java_version: number;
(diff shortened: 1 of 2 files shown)


rgrunber:
I haven't reviewed this thoroughly, but noticed the workflow seems to be :

Check if some JRE on the system is compatible -> If not, try the binary server

However, if the binary server fails (eg. not trusted, and we don't trust it), we seem to be trying Java once again :
https://github.com/redhat-developer/vscode-xml/blob/53f04ba280f92dcc27d78f4485b1772fd72a5d4b/src/server/serverStarter.ts#L53-L62

I would just make sure that works as expected.

datho7561:
I get an error popup that says:

Activating extension 'redhat.vscode-xml' failed: Failed to launch binary XML language server and no Java is installed.

src/server/requirements.ts
```diff
@@ -153,7 +155,7 @@ function checkJavaVersion(java_home: string): Promise<number> {
     cp.execFile(java_home + '/bin/java', ['-version'], {}, (error, stdout, stderr) => {
       const javaVersion = parseMajorVersion(stderr);
       if (javaVersion < 11) {
-        reject(openJDKDownload('Java 11 or more recent is required to run. Please download and install a recent Java runtime.'));
+        reject(openJDKDownload('The Java version specified is older that Java 11.'));
```
rgrunber, line 158: "older than"

rgrunber (approved):
Seems fine to me. The only odd thing I found was :

image

I simulated Java 8 being detected, and got one warning prompt regarding the version being insufficient, and the option to download java / always use binary server. The other prompt is from the natural flow of the program to attempt to use the binary server. Probably won't happen normally, since I am using a locally built/untrusted binary.

I guess there's nothing too wrong here because the first prompt is more informational, and the user need not make a decision there.

PR #1054, opened 2025-02-12
File associations should be unique by pattern, not system id

### LINKED ISSUE #1053: XMLFileAssociations: Comparing schemas but should compare files
In XMLFileAssociations (both add and remove) there is a lookup for existing schemas (systemId), instead there should be a lookup for existing files (or patterns)

Image

datho7561:
Fixes #1053


angelozerr (approved):

angelozerr:
@datho7561 if you think it is good please merge it. Thanks!

PR #1074, opened 2025-04-29
Add smoke tests for the binary

datho7561:
Often, we run into issues where the binary version of lemminx doesn't launch properly due to missing GraalVM reflection configuration. We can't catch these issues using the current lemminx tests, because those tests do not run against the fully assembled binary server.

This PR includes a smoke test suite for vscode-xml where I check the diagnostic messages of an invalid XML document against an XSD schema in order to make. These tests can be run locally or through CI, and default to using the Java version of lemminx (**TODO:** it would be nice if the tests were configured to use the binary by default).

I also include a GitHub Action that builds the binary server and then runs the smoke test suite against it. (vscode-xml should use the binary server regardless of the `settings.json` if the `.jar` is not available).

This suite is a modified copy of [vscode-lemminx-schematron](https://github.com/datho7561/lemminx-schematron/tree/main/vscode-schematron)'s test suite.


datho7561:
Looking into deleting the artifacts after the build so that we don't use up the storage quota.

PR #1078, opened 2025-04-30
Fix extension warnings

### LINKED ISSUE #98: Warnings in extension development

datho7561:
We were accessing some scoped settings incorrectly; this should access them correctly.

Fixes #98

### DIFF
--- a/src/client/tagClosing.ts
+++ b/src/client/tagClosing.ts
@@ -34,7 +34,7 @@ export function activateTagClosing(tagProvider: (document: TextDocument, positio
     if (!supportedLanguages[document.languageId]) {
       return;
     }
-    if (!workspace.getConfiguration(undefined, document.uri).get<boolean>(configName)) {
+    if (!workspace.getConfiguration().get<boolean>(configName)) {
       return;
     }
     isEnabled = true;
--- a/src/settings/settings.ts
+++ b/src/settings/settings.ts
@@ -107,9 +107,8 @@ function verifyVMArgs() {
 }
 
 function verifyAutoClosing() {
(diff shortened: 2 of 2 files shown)


datho7561:
To review this PR: check out the extension logs. After this change, there should be no "scoped settings" warnings emitted by vscode-xml. Also make sure auto tag closing works as expected.

Issue #916, opened 2023-07-11
Extension crashing on start after updating to latest version


siferati:
The extension stopped working after updating to the latest version today.
Seems something related with libc?

```
[Error - 3:12:13 PM] Restarting server failed
  Message: Pending response rejected since connection got disposed
  Code: -32097 
/home/tiago/.vscode/extensions/redhat.vscode-xml-0.26.0-linux-x64/server/lemminx-linux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /home/tiago/.vscode/extensions/redhat.vscode-xml-0.26.0-linux-x64/server/lemminx-linux)
/home/tiago/.vscode/extensions/redhat.vscode-xml-0.26.0-linux-x64/server/lemminx-linux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/tiago/.vscode/extensions/redhat.vscode-xml-0.26.0-linux-x64/server/lemminx-linux)
```

rgrunber:
Comparing the 2 binaries, it seems like the newer one requires GLIBC_2.34 for certain symbols that before were found on 2.2.5.

```diff
--- /dev/fd/63	2023-07-11 08:30:30.551127674 -0400
+++ /dev/fd/62	2023-07-11 08:30:30.552127692 -0400
@@ -1,5 +1,5 @@
+0000000000000000 (GLIBC_2.2.5) __cxa_finalize
 0000000000000000 (GLIBC_2.2.5) _environ
-(GLIBC_2.12) pthread_setname_np 
 (GLIBC_2.14) memcpy 
 (GLIBC_2.15) posix_spawn 
 (GLIBC_2.17) clock_gettime 
@@ -13,9 +13,6 @@
 (GLIBC_2.2.5) close 
 (GLIBC_2.2.5) closedir 
 (GLIBC_2.2.5) connect 
-(GLIBC_2.2.5) dlclose 
-(GLIBC_2.2.5) dlopen 
-(GLIBC_2.2.5) dlsym 
 (GLIBC_2.2.5) dup 
 (GLIBC_2.2.5) dup2 
 (GLIBC_2.2.5) __environ 
@@ -70,7 +67,6 @@
 (GLIBC_2.2.5) kill 
 (GLIBC_2.2.5) lchown 
 (GLIBC_2.2.5) __libc_current_sigrtmax 
-(GLIBC_2.2.5) __libc_start_main 
[...]

PR #1081, opened 2025-05-08
Use 'ubuntu-22.04' in native-image CI

datho7561:
20.04 was retired from GitHub Actions in April.


rgrunber (approved):

rgrunber:
Just be aware of https://github.com/redhat-developer/vscode-xml/pull/917 . Full explanation at https://github.com/redhat-developer/vscode-xml/issues/916#issuecomment-1630739626 . This might break older systems but not much we can do.

PR #1083, opened 2025-05-09
Drop support for Alpine Linux-specific version

datho7561:
We can't build a binary version of the language server for Alpine Linux any more due to bugs/limitations in GraalVM and changes to the runners in GitHub Actions.

It's _probably_ okay to drop Alpine Linux support, since Alpine Linux is usually used on servers instead of as a desktop operating system, and it's a real pain to set up as a desktop operating system.

Alpine Linux users should still be able to use the extension if they install Java.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Build
 
  * Add smoke tests for the binary server. See [#1075](https://github.com/redhat-developer/vscode-xml/issues/1075).
+ * Drop support for Alpine Linux-specific version. See [#1083](https://github.com/redhat-developer/vscode-xml/pull/1083)
 
 ## [0.28.1](https://github.com/redhat-developer/vscode-xml/milestone/40?closed=1) (April 9, 2025)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Build
 
  * Add smoke tests for the binary server. See [#1075](https://github.com/redhat-developer/vscode-xml/issues/1075).
+ * Do not build Alpine Linux-specific version. See [#1083](https://github.com/redhat-developer/vscode-xml/pull/1083)
```
rgrunber, line 12: Is it worth saying "Drop support for Alpine Linux-specific version." ?

datho7561: yeah

rgrunber (approved):

rgrunber:
https://github.com/redhat-developer/vscode-xml/blob/62f6a66ef58d5bdc3f7c9ecef26b5bb1042bc00c/src/server/binary/binaryServerStarter.ts#L209-L218 mentions the path to the alpine binary, but if that's gonna be the linux one, then it should probably be removed ?

PR #1100, opened 2025-07-31
Use Node 20

datho7561:
Node 18 is EOL and has been for a bit, so we should use Node 20 wherever possible.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,6 +6,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-node@v4
+      with:
+        node-version: '20'
     - name: Install dependencies
       run: npm i --also=dev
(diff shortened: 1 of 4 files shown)


datho7561:
I don't think we need to explicitly run setup nodejs. I think the runner comes with the latest LTS nodejs version available on the path (as evidenced by not needing to run it for the CI step). As a result, I've removed the setup-node step from the release job.

rgrunber:
Would it be https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md ? It seems to imply Node 20 is the default ?

rgrunber (approved):
I'm fine with this. I have a slight preference for listing the action (setup-node : 20) because it's more clear what version is being run against. However, given the limited time for things like this maintenance, maybe trying with the latest if possible saves a lot of headaches.

datho7561:
Sounds like a good idea, I'm convinced. I also locked the node version in the CI action I wrote.

Issue #930, opened 2023-09-06
Release for linux-aarch64


jamilraichouni:
Hi!

Would love to see a Linux arm64/ aarch64 build for lemminx. 
Is that somewhat on the agenda? 😀

Many thanks!
Jamil

angelozerr:
@rgrunber @datho7561 do you think it is easily doable?

rgrunber:
I assume this is referring to the lemminx binary, since it should be possible to install the generic vscode-xml extension on linux-aarch64. For the binary, the main issue would be that GitHub doesn't provide such runners : https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources .

angelozerr:
@rgrunber I think this issue can be closed?

angelozerr:
Ok thanks @datho7561 for your clarification!

jay-babu:
> Depends on [actions/runner-images#5631](https://github.com/actions/runner-images/issues/5631)

This has been closed as of 4 days ago. Is it possible now? @rgrunber @datho7561

rgrunber:
Yes (and GraalVM supports it), but based on https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/ they are not freely available for public repositories.

javalsai:
Despite not being supported natively by runners, it's easily doable with virtualization using an action such as https://github.com/uraimo/run-on-arch-action. It's used by several other projects to provide binaries for alternative architectures.

PR #1115, opened 2025-11-10
Smoke test all binaries using matrix build

datho7561:
This lets us smoke test the binaries for every platform instead of just linux x86_64.

The builds are run in parallel, which looks a bit messy in the output but hopefully will mean it takes around 10 minutes to get all the results.

I wrote the windows part so that it can be expanded to include arm64 in the future.

Also fixes a bug preventing the linux arm64 binary from being used.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,43 +16,42 @@ jobs:
     - name: Run ESLint
       run: npm run eslint
 
-  build-test-native-image:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-      with:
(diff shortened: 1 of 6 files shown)


vidhanio (approved):
LGTM

datho7561:
Fixes https://github.com/redhat-developer/vscode-xml/issues/1145

rgrunber (approved):
Seems reasonable. ~~Just curious, was the linux-arm64 support wrongly using lemminx-linux-x86_64 ?~~

**Update:** Nevermind, you explained here https://github.com/redhat-developer/vscode-xml/issues/930#issuecomment-3512919872 .

PR #1138, opened 2026-04-18
XML Minify support

### LINKED ISSUE #609: Add command "Minify XML"
It would be nice if we could minify XML files using this extension.
Use cases: send large unformatted files to a messaging queue server

Thank you!

angelozerr:
XML Minify support

Fixes #609


angelozerr:
Replaced with https://github.com/redhat-developer/vscode-xml/pull/1139

PR #1144, opened 2026-04-21
Update version of Intel macOS in release build

datho7561:
macOS 13 runners are no longer available. macos-latest-large points to the latest available Intel macOS


rgrunber (approved):
Makes sense. There's no more macos-13 according to https://github.com/actions/runner-images#available-images

PR #1158, opened 2026-07-21
Fix tests

datho7561:
- Add tsconfig so that we don't type check libraries (this wastes time and sometimes yields false positives).
- Bump setup-graalvm to get windows build working. There's a weird bug related to MSVC not being set up properly that should be fixed by a newer version of the action.
- Upgrade testing framework (@vscode/test-electron and @vscode/test-cli) to latest version to address a failure on macOS
- Use Node 22 in CI, since the test framework requires it


shin19991207 (changes requested):
I think we should update CI and release to Node 22? `@vscode/test-electron`@3.1.0 package declares node >=22 (https://github.com/microsoft/vscode-test/blob/003b260c457c1d037a564b4c730624ac45b7789a/package.json#L21-L24).

- https://github.com/datho7561/vscode-xml/blob/eda403996ae70a7131b410a4ffc858eae230df3e/.github/workflows/ci.yaml#L11
- https://github.com/datho7561/vscode-xml/blob/eda403996ae70a7131b410a4ffc858eae230df3e/.github/workflows/release.yml#L86
- https://github.com/datho7561/vscode-xml/blob/eda403996ae70a7131b410a4ffc858eae230df3e/.github/workflows/release.yml#L191

shin19991207:
There's another Node version 20 that's used here: https://github.com/datho7561/vscode-xml/blob/1b775300ee512fe159fbc6fd9fc966f278938e03/.github/workflows/release.yml#L86

datho7561:
Good catch, thanks!

shin19991207 (approved):
Looks good to me. Thanks, David!

PR #1166, opened 2026-09-11
Fix/xml encoding detection encoding

### LINKED ISSUE #759: Respect XML document encoding
An XML document can have a document encoding :

`<?xml version="1.0" encoding="UTF-8"?>`

We should detect (entirely on client-side) the type of encoding and set it appropriately. VS Code currently doesn't handle this, so maybe we can support it on our end for now.

angelozerr:
## Summary

Fixes #759

Detect the `encoding` attribute from the XML prolog (e.g. `<?xml version="1.0" encoding="ISO-8859-1"?>`) and automatically reopen the document with the correct encoding.

This encoding detection is applied:
- **On file open**: when an XML document is opened, the extension parses the prolog and reopens the file with the declared encoding if it differs from the current one.
- **On file save**: when an XML document is saved, the extension re-checks the prolog encoding, handling the case where the user changed the encoding declaration.

The feature uses the `TextDocument.encoding` API introduced in VS Code 1.100. On older versions, the detection is gracefully skipped (no error, no overhead).

### Demo

Here is a demo:

VscodeXMLEncoding

## Test plan

- [ ] Open an XML file with `encoding="ISO-8859-1"` -> verify the file is automatically reopened with the correct encoding
- [ ] Open an XML file with `encoding="UTF-8"` -> verify no reopen occurs
- [ ] Open an XML file without encoding attribute -> verify no reopen occurs
- [ ] Edit the prolog encoding and save ->  verify the file is reopened with the new encoding
- [ ] Open a non-XML file ->  verify no encoding detection occurs
- [ ] Run on VS Code < 1.100 -> verify no error occurs

### DIFF
--- a/src/client/xmlEncoding.ts
+++ b/src/client/xmlEncoding.ts
@@ -1,0 +1,145 @@
+/**
+ *  Copyright (c) 2026 Red Hat, Inc. and others.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-v20.html
+ *
+ *  Contributors:
+ *  Red Hat Inc. - initial API and implementation
+ */
+
+import { ExtensionContext, TextDocument, window, workspace } from 'vscode';
+
+const XML_TO_VSCODE_ENCODING = new Map<string, string>([
+  ['utf-8', 'utf8'],
+  ['us-ascii', 'utf8'],
+  ['ascii', 'utf8'],
+  ['utf-16', 'utf16le'],
+  ['utf-16le', 'utf16le'],
+  ['utf-16be', 'utf16be'],
+  ['iso-8859-1', 'iso88591'],
+  ['iso-8859-2', 'iso88592'],
+  ['iso-8859-3', 'iso88593'],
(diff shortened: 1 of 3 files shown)


angelozerr:
@fbricon @datho7561 please review the PR.

datho7561:
When opening a file with an encoding other than utf-8, the encoding isn't updated in the bottom right until I make changes in the file.

angelozerr:
It should, which vscode version do you have?

datho7561:
The encoding API exists and changing the encoding through the API works; it seems like there's another callback that's resetting the encoding to UTF-8 around a second after I open the document.

datho7561:
1.137.0

datho7561:
Try adding a breakpoint into the `checkAndFixEncoding` function, then closing and opening the document a few times. It only seems to work the first time you open the file for some reason.

datho7561:
It works more often now, probably around half the time, but it still doesn't work consistently. I think it might be a bug in VS Code iteself, and it might be Linux specific.

datho7561 (approved):
I think we can look into improving it in the future, but for now, this is better than no detection of the encoding whatsoever.

datho7561:
Angelo pointed out to me that editorconfig allows configuring the encoding, so in my workspace vscode-xml and editorconfig were competing to set the file encoding of the file.

